PR #17, opened 2018-03-20
Pr wfmdata variable size

xresende:
- change pwrsupply classes to allow for length-varying wfmdata setpoints.
- configuration comparisons in servconf will have to be generalized!

### DIFF
--- a/siriuspy/siriuspy/pwrsupply/model.py
+++ b/siriuspy/siriuspy/pwrsupply/model.py
@@ -199,7 +199,10 @@ class PowerSupply(_PSCommInterface):
         return self._controller.write('WfmLabel-SP', value)
 
     def _set_wfmdata(self, value):
-        value = [v for v in range(min(len(value, _max_wfmsize)))]
+        if isinstance(value, (int, float)):
+            value = [value, ]
+        elif len(value) > _max_wfmsize:
+            value = value[:_max_wfmsize]
         self._setpoints['WfmData-SP']['value'] = value
         return self._controller.write('WfmData-SP', value)
 
--- a/siriuspy/siriuspy/pwrsupply/model.py
+++ b/siriuspy/siriuspy/pwrsupply/model.py
@@ -10,6 +10,7 @@ from epics import PV as _PV
 
 from siriuspy.namesys import SiriusPVName as _SiriusPVName
 from siriuspy.envars import vaca_prefix as _VACA_PREFIX
+from siriuspy.csdevice.pwrsupply import max_wfmsize as _max_wfmsize
 from siriuspy.factory import NormalizerFactory as _NormalizerFactory
 from siriuspy.epics import connection_timeout as _connection_timeout
 from siriuspy.epics.computed_pv import QueueThread as _QueueThread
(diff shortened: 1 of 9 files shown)


siriuspy/siriuspy/pwrsupply/model.py
```diff
@@ -197,16 +199,8 @@ def _set_wfmlabel(self, value):
...
-        n = len(self._setpoints['WfmData-SP']['value'])
-        if isinstance(value, (int, float)):
-            self._setpoints['WfmData-SP']['value'][0] = value
-        elif len(value) == n:
-            self._setpoints['WfmData-SP']['value'] = value
-        else:
-            for i in range(min(len(value), n)):
-                self._setpoints['WfmData-SP']['value'][i] = value[i]
-        value = self._setpoints['WfmData-SP']['value']
+        value = [v for v in range(min(len(value, _max_wfmsize)))]
```
fernandohds564: bug here.

siriuspy/siriuspy/pwrsupply/pru.py
```diff
@@ -212,11 +214,16 @@ def get_connected(self, ID_device):
...
+
     def set_wfmdata(self, ID_device, wfmdata):
         """Set waveform of a device."""
         sorted_IDs = sorted(self._waveforms.keys())
         if ID_device not in sorted_IDs:
-            raise ValueError
+            print('ID_device {} not defined!'.format(ID_device))
+            return
         else:
             self._waveforms[ID_device] = wfmdata[:]
```
fernandohds564: what if wfmdata do not have the right length?
If its length is smaller than 4000, all values will be overwritten, right?
If its length is larger than 4000, the resulting waveform will also be larger.

siriuspy/siriuspy/servconf/types/rmp_bo_ps.py
```diff
@@ -1,10 +1,10 @@
 """Booster ramp configuration definition."""
 import copy as _copy
-from siriuspy.csdevice.pwrsupply import default_wfmsize as \
-    _default_wfmsize
+from siriuspy.csdevice.pwrsupply import max_wfmsize as \
+    _max_wfmsize
```
fernandohds564: The same import in all the other files has only one line. Why use 2 lines here?

siriuspy/siriuspy/pwrsupply/model.py
xresende: good catch. fixing it...

siriuspy/siriuspy/pwrsupply/pru.py
xresende: PRU library deals with variable-size wavelengths. PRU._waveforms is passed as argument to a function in this library. Since class PRU represents PRU and its library, I decided not to impose policies on wfmdata in it, just like the PRU library. The other classes that uses PRU do proper checkings. (model.PowerSupply, in this case)

siriuspy/siriuspy/servconf/types/rmp_bo_ps.py
xresende: it had to be a two-liner when the imported variable was named '_default_wfmsize'.  text auto replace in atom was not that smart! :smiley:. changing it...

anacso17:
OK

PR #18, opened 2018-03-20
Pr add firmware version controller

xresende:
- added readout of firmware_version from power supply controller (simulated and real)
- updated PS interlock labels


siriuspy/siriuspy/csdevice/pwrsupply.py
```diff
@@ -39,11 +39,12 @@
...
-                     'Undervoltage on DC-Link', 'DC-Link input relay fail',
-                     'DC-Link input fuse fail', 'Fail on module drivers',
-                     'Reserved', 'Reserved',
-                     'Reserved', 'Reserved', 'Reserved', 'Reserved',
+ps_hard_interlock = ('Falha na fonte 1', 'Falha na fonte 2',
+                     'Falha na fonte 3', 'Sensor de fumaça',
+                     'Interlock externo', 'Sobre-tensão na fonte 1',
+                     'Sobre-tensão na fonte 2', 'Sobre-tensão na fonte 3',
+                     'Sub-tensão na fonte 1', 'Sub-tensão na fonte 2',
+                     'Sub-tensão na fonte 3', 'Reserved',
```
fernandohds564: Are you sure about the language?

xresende: yes, i deliberately switched back to portuguese. this is due to the fact that now the primary source of specification of power supply comm. interface is exposed from our sirius wiki pages. there, in the excel table, the interlock names are still in portuguese. i will request ELP staff to update it to english, since this is our language choice for the control-system. when the spec is updated to english i will update <code>siriuspy</siriuspy> back to english accordingly.

xresende: just a way to propagate to the control-system the pending issue of translation to english the PS spec, thus reinforcing the need to eventually do it.

PR #35, opened 2018-03-30
Pr thread module

fernandohds564:
Hi guys, I took a look at the Class `QueueThread` defined in module `computed_pv` and fount it very interesting!!

I plan to use this Class in the `timesys` package, so I have generalized it a little bit and changed its location to the new module `thread`, where I also added a Class that I had already been using in `timesys` module from `machine-applications` repo.


fernandohds564:
Yes @ximenes. It is useful.
All oneline statements in python are thread safe. That is why a Lock is not needed.

PR #37, opened 2018-04-03
Packages to interact with Timing System

fernandohds564:
News from timing system branch.


fernandohds564:
Very good @xresende! thanks for the corrections.

Do I implement the tests now?

PR #39, opened 2018-04-05
Add classes and functions to deal with SigGen configurations

xresende:
- a file <code>siggen-configuration.txt</code> was added to the web service.
- created a SigGenConfig class in <code>pwrsupply/siggen.py</code>.
- added methods in <code>servweb</code> and <code>PSSearch</code> to deal with siggen configurations
- added corresponding unittests.


siriuspy/siriuspy/pwrsupply/siggen.py
```diff
@@ -0,0 +1,159 @@
...
+        """Return list with signal waveform."""
+        t = _np.linspace(0.0, self.duration, nr_points)
+        if self.type in ('Sine', 'DampedSine'):
+            # TODO: confirm!
+            if self.type == 'Sine':
+                amp = self.amplitude * _np.ones(t.shape)
+            else:
+                amp = self.amplitude * _np.exp(-t/self.decay_time)
+            wfm = self.offset * _np.ones(t.shape)
+            phase = self.freq*t % (2*_np.pi)
```
fernandohds564: I think there is a bug here... the interpretation of `self.freq` is the frequency, not the angular frequency.

I think the correct code would be:

`phase = (2 * _np.pi) * (self.freq * t % 1)`

xresende: you're right, @fernandohds564 . thanks!

PR #45, opened 2018-04-10
Creation of Package search

fernandohds564:
Summary:
 - Creation of subpackage `search` with modules:
    - `ps_search`;
    - `ma_search`;
    - `hl_time_search` (old `timesys/hl_types_data`);
    - `ll_time_search` (merge of timesys `IOs` and `Connections` classes);
 - Include High Level database in `csdevice/timesys` module.
 - Creation of tests for subpackage search and update of `csdevice/timesys` tests;


siriuspy/siriuspy/csdevice/timesys.py
```diff
@@ -48,10 +48,28 @@
     )
 
 
+class Constants:
```
xresende: Can we use <code>Const</code> instead of <code>Constants</code>, as in other modules?

PR #50, opened 2018-04-12
Configure Log File

fernandohds564:
Hi guys! I created a function called `configure_log_file` in module `util` to setup the `logging` module for us.

I also changed the function `print_ioc_banner` to use the logging module instead of the print function.

If this PR is accepted we need to invoke the function `configure_log_file` in the beggining of the function `run` of our IOCs to setup the logger, than we can use the module along all the files of the subpackage and even in different threads of our application.


fernandohds564:
A PR (https://github.com/lnls-sirius/machine-applications/pull/23) was created in `machine_applications` repository adding the call to `configure_log_file` to **All IOCs**.

PR #58, opened 2018-05-02
Migrate databases to csdevice

anacso17:
This PR migrates some databases from machine-application to siriuspy.csdevice.
It creates  `currinfo.py`, `opticscorr.py` and `posang.py` modules to csdevice and adds their respective tests.


siriuspy/siriuspy/csdevice/currinfo.py
```diff
@@ -0,0 +1,91 @@
...
+        setattr(obj, const, i)
+
+
+Const._init()  # create class constants
+
+
+def get_charge_database():
+    """Return CurrentInfo-Charge Soft IOC database."""
+    pvs_database = {
+        'Version-Cte':        {'type': 'string', 'value': 'UNDEF'},
```
xresende: @anacso17 , we have been using <code>'str'</code> for pcaspy string PV type in database entries, whereas you used <code>'string'</code>. do you know if both options are available? maybe we should try to stick with one option.

xresende: update: PCASpy documentation mention <code>'string'</code> as an option for the PV type, but does not mention <code>'str'</code>. i will create another PR replacing all <code>'str'</code> instances to <code>'string'</code>.

xresende:
@anacso17 , PR merged. you can delete branch if it will not be used any longer.

PR #81, opened 2018-08-21
SOFB updates

fernandohds564:
Guys, this PR is a little bit large and brings small changes to several sub-packages. The main idea was to update the SOFB code to make its implementation closer to the standard developed for the other IOCs. In this process other packages were also updated, such as timing and BPMs.

The SOFB code is not completely debugged yet, but I will add these changes in a future PR.


fernandohds564:
@gu1lhermelp, I will merge this one, ok?
If you find any bug or have any suggestions, we can review them in later PRs.

PR #86, opened 2018-09-25
Change default kick units of correctors and pm.

xresende:


xresende:
Please check if other changes (IOCs | HLAs) are needed as a consequence of this modification.

PR #110, opened 2018-11-22
New Const and EnumTypes in csdevice modules

xresende:
* Define a general ETypes in `csdevice.util`
* Create local `ETypes` derived from global one.
* Standardize how `Const` is defined in `csdevice` modules


xresende:
@fernandohds564 , last commit https://github.com/lnls-sirius/dev-packages/pull/110/commits/622795a8590520f2dc4096b989bc8bb49480db94 is a major refactoring of `csdevice.orbitcorr`, changing how module constants are stored in classes. 

I have checked these changes by creating objects of the classes in the module and they ssem to work. But please check if you agree with these modifications... The idea is to make the module as close as possible to the other modules in `csdevice`.

the package `as-ap-sofb` will need minor changes in order to meet changes in `csdevice.orbitcorr`. I will submit a PR in `machine-applications`

fernandohds564:
@xresende, I've made a small refactoring to the `csdevice.orbitcorr` module. If you agree with these changes, you can merge the PR.

PR #122, opened 2018-12-15
Timing updates

fernandohds564:
This PR brings the refactoring I have made to the HLTiming code.

It is worth mention that this PR introduces `pcaspy` as a requirement for `siriuspy`.


siriuspy/siriuspy/csdevice/timesys.py
```diff
@@ -469,9 +475,9 @@ def get_evg_database(prefix=None, only_evg=False):
     if only_evg:
         return db
 
-    for clc in clocks_ll2hl_map.keys():
+    for clc in Const.ClkLL2LLMap.keys():
```
anacso17: Is this name (Const.ClkLL2LLMap) right? Just checking, because there is a difference between the old and the new names meaning.

fernandohds564: Thanks @anacso17! It was wrong. The name was supposed to be `ClkLL2HLMap`.

PR #133, opened 2019-02-21
Fix FAC ACDC power state behaviour

gu1lhermelp:


gu1lhermelp:
Tested with simulated IOCs

PR #153, opened 2019-03-14
ENH: Add repo scripts to the list of sirius repos.

fernandohds564:


xresende:
@fernandohds564 , I thought we were not using this list anymore...

fernandohds564:
@xresende, the script https://github.com/lnls-fac/scripts/blob/master/bin/sirius-lnls-gitall.py uses this list.
However, it can be changed instead.
I think it's better to remove this list if we want not to use it any more.

xresende:
I think [sirius-lnls-gitall.py](https://github.com/lnls-fac/scripts/blob/master/bin/sirius-lnls-gitall.py ) is to be deprecated soon, now that we will use ansible.

xresende:
We can postpone the decision of deleting the list in <code>siriuspy.envars</code> since you may still be using [https://github.com/lnls-fac/scripts/blob/master/bin/sirius-lnls-gitall.py](https://github.com/lnls-fac/scripts/blob/master/bin/sirius-lnls-gitall.py)

PR #155, opened 2019-03-15
Add bsmp_init_group when invalid_id response is received

gu1lhermelp:
This PR fixes the problem that occurs when the IOC is running, and someone sends the BSMP command `reset_udc`. 

In this command the BSMP entities will be reset, causing all groups to be removed. This leaves the IOC in a states where it continues to read the groups but they don't exist causing the PVs to be disconnected and not being updted.

This PR adds an ack check and create the groups again in case in receives an `invalid_id` response.

### DIFF
--- a/siriuspy/siriuspy/pwrsupply/prucontroller.py
+++ b/siriuspy/siriuspy/pwrsupply/prucontroller.py
@@ -1042,6 +1042,9 @@ class PRUController:
                     copy_var_vals[id][self._params.ConstBSMP.V_I_LOAD2] += \
                         0.00001*_random.uniform(-1.0, +1.0)
 
+            elif ack[id] == _Response.invalid_id:
+                self._connected[id] = False
+                self._bsmp_init_groups()
             else:
                 self._connected[id] = False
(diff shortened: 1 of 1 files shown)


siriuspy/siriuspy/pwrsupply/prucontroller.py
```diff
@@ -1042,10 +1043,15 @@ def _bsmp_update_variables(self, device_ids, group_id):
...
 
+            elif ack[id] == _Response.invalid_id:
+                self._connected[id] = False
+                create_groups = True
             else:
                 self._connected[id] = False
         # processing time up to this point: 19.4 ms @ BBB1
         # print('time3: ', _time.time() - t0)
+        if create_groups:
+            self._bsmp_init_groups()
```
xresende: @gu1lhermelp , do we need to recreate variables groups for all BSMP devices? maybe *create_groups* should be turned into a list of BSMP IDs for which variables groups need to be recreated and then passed as argument for a generalized <code>self._bsmp_init_groups(list_ids=None)</code>

PR #193, opened 2019-05-03
Pr remove ps rampramp

xresende:
### Remove ramp of ramp functionality in PS iocs.

TODO before merging:
- **!!! Needs additional checking  (Ramp subpackage, Ramp HLA, MA and PS HLA)**


xresende:
@fernandohds564 and @anacso17, I need help assessing whether moving *OpMode-Sel* to the top of *global_config*, as it should, does not break loading/saving existing configurations. what you do think?

siriuspy/siriuspy/servconf/types/global_config.py
```diff
@@ -931,6 +909,29 @@ def get_dict():
     ['TB-04:PS-CV-1:OpMode-Sel', _off, 0.0],
     ['TB-04:PS-CV-2:OpMode-Sel', _off, 0.0],
 
+    ['TB-Fam:PS-B:PwrState-Sel', _off, 0.0],
```
fernandohds564: @xresende, Isn't it a problem to put `PwrState-Sel` after `OpMode-Sel` for the cases where we are turning on the power supplies and setting their modes to something different then `SlowRef`?

Maybe it would be better to remove `PwrState-Sel` from this configuration. Maybe we could've another configuration type to turn `On` and `Off` all power supplies or make some script  to do it...

xresende: @fernandohds564 , your are right. if both `PwrState` and  `OpMode` are in the configuration, in some cases `PwrState` has to preceed `OpMode` and in other cases the contrary, as I defined in this PR,  has to happen (for example, when we want to set parameters of the power supply, regardless of what operation mode it was previously). 

We should move `PwrState-Sel` to another configuration type and adapt HLA clients to load `global_config` only afetr  devices are turned on...

PR #197, opened 2019-05-11
Power Supply IOC Classes Revision

xresende:
Major modification of PS IOC classes

The motivation was to improve behavior of the power supply ioc in the transient period right after a PV is written, may it be power on state, ps operation mode or current setpoint. These inconsistent transient states of the ioc were making it difficult for some HLA clients (configuration loader, magnet cycling application) to properly function with reasonable intentional sleeps between write and read operations.

Main modifications:
- values of ioc write requestes are now accepted into epics database right away. 
- these write requests are queued and processed in chronological order by power supplies.
- In bbb scan, setpoint PVs other than *-Cmd are not read from `BeagleBone` to epics db.
- minor modification of `BeagleBone` class interface
- In `PRUController` setting an operation mode already in use is processed just as if the PRU was in another mode. Previously nothing was done. This change is important for cases of beaglebones connected to more than one UDC.

Tests done with this PR:
- Load config with (TB quads, TB corrs, TB dipoles): OK
- Automated MA Cycling with (TB quads, TB corrs, TB dipoles): OK
- Manual MA Cycling with (TB quads, TB corrs, TB dipoles): Sometimes fails. delays? MA?

Update 2019-07-19:

- Tested this version in all power supplies with exception of TB quadrupoles and correctors.
- All cyclings within the HLA ps-cycle worked! The same for load config.
- TB correctors and quadrupoles still should use `deploy_2018_05_07`.
- PR is ready for merge!


xresende:
@anacso17 , please read the PR description that was updated. This PR is ready for merge!

PR #239, opened 2019-08-19
Update psmodel FBP

xresende:
- Add two more PVs: SwitchesTemperature-Mon and PWMDutyCycle-Mon


xresende:
**tested OK with real IOCs**

PR #248, opened 2019-08-20
Update psmodel FAP_2P2S

xresende:


xresende:
**Real IOC testes ok**
no serial connection to second power supply.

PR #274, opened 2019-10-05
Update PS IOCS FBP data for new wfmref firmware

xresende:
Update PS IOC to work with current BSMP spec but driving ramps still iusing beaglebones.

tested:

- BO FBPs (config load, ps test, cycle)


xresende:
PR deprecated!

PR #314, opened 2019-11-18
MA.MNT: change SI Dipole Normalizer to consider only B1B2 currents in energy calculation

anacso17:


anacso17:
@xresende I will merge this PR for now, but when you return we can discuss more about the issue.

PR #352, opened 2020-01-16
Fix bugs

anacso17:


siriuspy/siriuspy/csdevice/posang.py
```diff
@@ -1,23 +1,18 @@
 """Define PVs, contants and properties of PosAng SoftIOCs."""
 
 from siriuspy.csdevice import util as _cutil
-from siriuspy.namesys import SiriusPVName as _PVName
+# from siriuspy.namesys import SiriusPVName as _PVName
```
fernandohds564: @anacso17, I think it is better to remove this line instead of commenting it out...

PR #384, opened 2020-03-02
Add redundancy web and configdb servers

xresende:
as requested by @edupcoelho ...


fernandohds564:
@xresende, are going to do the same for the `ConfigDBClient` class in this PR or should we leave that one for another PR?

xresende:
> @xresende, are you going to do the same for the `ConfigDBClient` class in this PR or should we leave that one for another PR?

@fernandohds564 , I just pushed a commit with it. Please take a look.

PR #385, opened 2020-03-03
Replace WFMSIZE constants

xresende:


siriuspy/siriuspy/ramp/waveform.py
```diff
@@ -699,7 +701,7 @@ class Waveform(_WaveformMagnet):
     """Waveform class for general magnets."""
 
     def __init__(self, psname, dipole=None, family=None, strengths=None,
-                 currents=None, wfm_nrpoints=_DEF_WFMSIZE):
+                 currents=None, wfm_nrpoints=_DEF_WFMSIZE_FBP):
```
anacso17: I think this should be `_DEF_WFMSIZE_OTHERS` instead of `_DEF_WFMSIZE_FBP`

xresende: I think this is a dubious case. I chose _FBP_ simply because there are more corrector power supplies than family power supplies. Do you see particular reasons in favor of _OTHERS_ ?

anacso17: I think _FBP_ is a particular case, just like the variable names say. So, if the user uses this class to build a corrector waveform, the particularity of this parameter already have to be remembered.

xresende: Ok, I have changed the default argument in the constructor.

PR #397, opened 2020-03-10
Add config type 'si_bbbproc'

xresende:


xresende:
@lerwys , please check if PV names are ok. Also, what about _RENABLE_ for **BbBProc-V** **BbBProc-L** ?

PR #400, opened 2020-03-13
Update ClientConfigDB.get_config_types

xresende:
so that it get list from templates, not from database

### DIFF
--- a/siriuspy/siriuspy/clientconfigdb/configdb_client.py
+++ b/siriuspy/siriuspy/clientconfigdb/configdb_client.py
@@ -65,7 +65,7 @@ class ConfigDBClient:
 
     def get_config_types(self):
         """Get configuration types existing as database entries."""
-        return self._request()
+        return self._make_request()
 
     @staticmethod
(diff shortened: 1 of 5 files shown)


siriuspy/siriuspy/clientconfigdb/configdb_client.py
```diff
@@ -64,8 +64,13 @@ def get_nrconfigs(self):
         return self._make_request(stats=True)['count']
 
     def get_config_types(self):
-        """Get all configuration types."""
-        return self._make_request()
+        """Get configuration types existing as database entries."""
+        return self._request()
```
fernandohds564: @xresende, don't you have to use `self._make_request` here?

xresende: sure! I must have been too quick (and wrong) in using ms code's autocompletion... thanks.

xresende: fixed it.

PR #407, opened 2020-03-20
Pr add devices energy

xresende:


siriuspy/siriuspy/devices/energy.py
```diff
@@ -0,0 +1,106 @@
...
+from .device import DeviceApp as _DeviceApp
+
+
+class Energy(_DeviceApp):
+    """Energy devices."""
+
+    class DEVICES:
+        """."""
+
+        TB = 'TB'
```
fernandohds564: @xresende, I think it is better to use `namedtuple` here.

xresende: @fernandohds564 , I was waiting for you to submit a PR with your namedtuple implementation in order to convert in all device classes. By looking at your example (same v variable names, structures, etc), we could enforce standardization.

fernandohds564: ok! I think it will take sometime for that branch of mine become PRable.
Let's follow your strategy and standardize later.

PR #408, opened 2020-03-21
Empty siriuspy.csdevice

xresende:
In this PR I started moving device epics databases from _csdevice_ to the corresponding device subpackage:

- _siriuspy.csdevice.util_ was moved to _siriuspy.csdev_
- _siriuspy.csdevice.pwrsupply_ was moved to _siriuspy.pwrsupply.csdev_
- _siriuspy.csdevice.timesys_ was moved to _siriuspy.timesys.csdev_


xresende:
> Very nice! But I think it is better to move all other modules before merging this PR.

We could do it like that. But it simplifies things for me if this PR could be merged as soon as reviewers agree on what and how it implements the emptying of _csdevices_ sub-package, as we discussed days ago. I say that because I am working on other PR that presupposes this new refactoring. This current PR I wrote so that module migration from _csdevice_ to other sub-packages could be done gradually, in other _PR-empty-csdevice-..._ PRs... Unless someone takes the task (which definitely is not high priority) I antecipate it will take quite sometime for other modules to migrate out of _csdevice_.... In the meantime I would have to periodically merge the PR into the one I am wroking right now in order to avoid the possibility of a more complex conflicting merge at the end.

I think the title I gave to this PR, **Empty siriuspy.csdevice** might have misguided you. I am changing it. If you guys agree on mergeing it right away I should change it more appropriately to **Move siriuspy.csdevice.pwrsuppy**...

xresende:
@anacso17 and @fernandohds564 : my suggestion is to use the name _csdev.py_ for the module inside the sub-package to where the corresponding _csdevice_ dev database module will migrate. if you have another name in mind, let us discuss so that we could all use the same name.

PR #415, opened 2020-03-25
Move sofb 2 dev

fernandohds564:
merge only after #408 is merged.

PR #416, opened 2020-03-25
Move diag from csdevice

fernandohds564:
to be merged only after #415.


siriuspy/siriuspy/devices/sofb.py
```diff
@@ -2,7 +2,7 @@
 
 import time as _time
 
-from ..csdevice.orbitcorr import SOFBFactory
+from ..sofb.csdev import SOFBFactory
```
xresende: see https://github.com/lnls-sirius/dev-packages/pull/415#discussion_r398070390

PR #417, opened 2020-03-25
General siriuspy code standardization and cleanup

xresende:


siriuspy/siriuspy/cycle/conn.py
```diff
@@ -588,14 +619,14 @@ def check_current_zero(self, wait=5):
         """Return wether power supply PS current is zero."""
         return _pv_timed_get(self['Current-Mon'], 0, abs_tol=0.1, wait=wait)
 
-    def prepare(self, mode):
+    def prepare(self):
```
anacso17: this can cause a problem, once the classes that use this LinacPSCycler presupposes this interface

siriuspy/siriuspy/cycle/conn.py
```diff
@@ -588,14 +619,14 @@ def check_current_zero(self, wait=5):
...
         """Config power supply to cycling mode."""
         status = True
         if not self.check_current_zero(wait=0.5):
             status &= self.set_current_zero()
         return status
 
-    def is_prepared(self, mode):
+    def is_prepared(self):
```
anacso17: same here

siriuspy/siriuspy/cycle/conn.py
xresende: yes, I noticed that, reverted this code lines back to the original and commited. I am confused. In my working branch _PR-code_standardization_ these methods all have the original _mode_ argument. Did you comment on an intermediary commit?

siriuspy/siriuspy/cycle/conn.py
xresende: same comment here.

siriuspy/siriuspy/cycle/conn.py
xresende: oh. probably you commented right afeetr I created the first commit. Anyway, I fixed the changes.

xresende: I was confused at first... I noticed what you said and thought I had reverted back all deletes _mode_ arguments. I reverted for a few methods only. fixing it...

xresende: I run grep on _hla_, _machine-applications_ and _dev-packages_ for _LinacPSCycler_ repos.
I could only find the line

`siriuspy/cycle/main.py:                    self.cyclers[name] = LinacPSCycler(name)`

do we really need the method argument _mode_ for this specialization of _PSCycle_?

xresende: I commit a fix. I added a throwback code line in order to avoid pylint messages.
see
https://stackoverflow.com/questions/5893163/what-is-the-purpose-of-the-single-underscore-variable-in-python

I timed this throwback code and it takes roughly the same execution time as `pass`. we could you it in these situations.

anacso17: Nice! I will use this solution in these situations!

PR #418, opened 2020-03-26
Upgrade SOFB

fernandohds564:
Implement control of non-linear correction of measured orbit by application of 9th order bivariate polynome to the raw data.

### DIFF
--- a/siriuspy/siriuspy/sofb/bpms.py
+++ b/siriuspy/siriuspy/sofb/bpms.py
@@ -14,8 +14,10 @@ TIMEOUT = 0.05
 
 
 class BPM(_BaseTimingConfig):
+    """."""
 
     def __init__(self, name, callback=None):
+        """."""
         super().__init__(name[:2], callback)
         self._name = name
         self.ORB_CONV = self._csorb.ORBIT_CONVERSION_UNIT
@@ -137,15 +139,18 @@ class BPM(_BaseTimingConfig):
             k: _PV(LL_PREF+self.name+':'+v, **opt) for k, v in pvs.items()}
 
     def set_auto_monitor(self, boo):
+        """."""
         self._posx.set_auto_monitor(boo)
         self._posy.set_auto_monitor(boo)
 
     @property
     def name(self):
+        """."""
         return self._name
 
     @property
     def connected(self):
+        """."""
         conn = super().connected
         pvs = (
             self._posx, self._posy,
@@ -162,6 +167,7 @@ class BPM(_BaseTimingConfig):
 
     @property
     def is_ok(self):
+        """."""
         if not super().is_ok:
             return False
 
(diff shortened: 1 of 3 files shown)


siriuspy/siriuspy/sofb/bpms.py
```diff
@@ -621,22 +743,57 @@ def calc_sp_multiturn_pos(self, **kwargs):
             v = v.reshape(-1, wsize)[:, mask]
             vs[a] = _np.std(v, axis=1)
 
-        s1, s2 = vs['A'] + vs['B'], vs['D'] + vs['C']
+        s1, s2 = vs['A'] + vs['C'], vs['D'] + vs['B']
```
xresende: These short-named variables do not comply with pylint. We have to discuss. I think we should continue using pylint but in some cases, like variable names in specific mathematical expressions, long names do not help, short names sometimes are even preferably from a mathematical standpoint. Maybe we can adopt one of the solutions:

- Turn off this pylint rule (A pitty, since in other situations I think having longer variable names help code readability)
- In my code setup, prepending an underscore seems to circumvent this rule. We could use this judiciously, for math expressions where the meaning of a variable is explicit by the context.

take a look at
https://stackoverflow.com/questions/21833872/why-does-pylint-object-to-single-character-variable-names

siriuspy/siriuspy/sofb/orbit.py
```diff
@@ -889,6 +947,7 @@ def _update_bpmoffsets(self):
 
     @staticmethod
     def _find_new_downsample(N, d, onlyup=False):
```
xresende: same discussion here, three-letter rule.

siriuspy/siriuspy/sofb/bpms.py
fernandohds564: I'll change these variable names. I think it's easier now, but I see and agree with your point. Maybe we have a third option, which is to configure pylint to do not bother us for some specific variable names (2nd answer of the stackoverflow question you posted).

siriuspy/siriuspy/sofb/orbit.py
fernandohds564: done!

siriuspy/siriuspy/sofb/bpms.py
fernandohds564: done!

xresende: on this issue, @murilobalves sent me yesterday a link to a ms code module that we could use to syncronize editor settings...

fernandohds564: interesting!

siriuspy/siriuspy/sofb/orbit.py
```diff
@@ -673,25 +673,25 @@ def _update_time_vector(self, delay=None, duration=None, channel=None):
...
             return
-        dl = (delay or self.timing.delay or 0.0) * 1e-6  # from us to s
+        dly = (delay or self.timing.delay or 0.0) * 1e-6  # from us to s
         dur = (duration or self.timing.duration or 0.0) * 1e-6  # from us to s
         channel = channel or self.bpms[0].acq_type or 0
 
         # revolution period in s
         if channel == _csbpm.AcqChan.Monit1:
-            dt = self.bpms[0].monit1period
+            dtim = self.bpms[0].monit1period
```
xresende: you think _dtim_ is better than _dtime_ I used the the standardization PR? if so, I can change all other places and use it. I have no strong opinion about this, other than trying to use the name name.

fernandohds564: `dtime` is much better! I'll change it!

fernandohds564: done!

PR #420, opened 2020-03-27
Create PosAng, OpticsCorr, PSDiag and CurrInfo subpackages

anacso17:

PR #424, opened 2020-03-27
Create PosAng, CurrInfo, OpticsCorr and PSDiag subpackages

anacso17:
Guys, this PR substitutes #420 removing from history those commits that are not related to PosAng, CurrInfo, OpticsCorr and PSDiag IOCs


siriuspy/siriuspy/currinfo/lifetime/main.py
```diff
@@ -0,0 +1,166 @@
...
+        self.run_callbacks('BuffSizeTot'+lt_type+'-Mon', len(val_abs_dqorg))
+
+    @staticmethod
+    def _least_squares_fit(timestamp, value, fit='exp'):
+        if fit == 'exp':
+            try:
+                value = _np.log(value)
+            except Exception:
+                return 0.0
+        n = len(timestamp)
```
xresende: @anacso17 , we are discussing how to deal with pylint pedantic complaints on three-letter variable name rule. We would like to rely on pylint to improve the code but having it listing these type of errors is a noiance. While we do not implement a definite solution maybe you can change the variable names. One simple solution is to use a prefix underscore.

See and contribute to the discussion I created for us in 
https://gitter.im/lnls-fac/community#

anacso17: I renamed the variable!

siriuspy/tests/currinfo/test_main.py
```diff
@@ -0,0 +1,55 @@
+# #!/usr/bin/env python-sirius
```
fernandohds564: Is it really worthy to keep a whole file commented out like this? With git tools it is easy to retrieve code whenever we want...
Couldn't we remove this file?

anacso17: I tried to fix these tests instead of commenting them, but I think I will remove them momentarily. In another opportunity I fix all new tests that I brought from machine-applications.

PR #425, opened 2020-03-27
Change SOFB variable names

fernandohds564:


siriuspy/tests/test_util.py
```diff
@@ -4,12 +4,13 @@
...
 import logging
-import numpy as np
 from unittest import mock, TestCase
+import numpy as np
+
 import siriuspy.util as util
 
 
-public_interface = (
+PUB_INTERFACE = (
```
xresende: Should I convert `PUBLIC_INTERFACE` in all modules to simpler `PUB_INTERFACE`? Use of editors find-and-replace is much simpler if we keep standard as much as possible.

fernandohds564: I think it's a good idea. but let me do it, I want to use `sed`! :joy:

fernandohds564: done!

xresende: nice. _grep_, _sed_, _code_... you choose your arm!

siriuspy/tests/ramp/test_waveform.py
```diff
@@ -222,29 +218,27 @@ def test_update(self):
...
         """Test duration."""
         # TODO: implement it!
 
-
     def _test_update(self):
         """Test update."""
         # TODO: implement using mock to substitute server
         bo_b = WaveformDipole()
-        defoc = Waveform(maname='BO-Fam:MA-QD', dipole=bo_b)
+        defoc = Waveform(psname='BO-Fam:MA-QD', dipole=bo_b)
```
xresende: @fernandohds564 , sure about this change?

fernandohds564: @anacso17 ?

anacso17: This is a right fix!

xresende: @anacso17 , it was not that. I found strange that the name of the `Waveform` constructor argument is `psname` and the string passed is of _MA_ discipline. I guess this a side effect of abandoning MA IOCs. We will eventually finalize refactoring of classes that used _ maname_...

xresende: @anacso17 , after you read my response, can you please resolve this conversation?

anacso17: Oh, my mistake, the discipline is really wrong!

PR #426, opened 2020-03-28
Standardize and Cleanup (Cont.)

xresende:
- Standardize throwaway comment.

### DIFF
--- a/siriuspy/siriuspy/cycle/conn.py
+++ b/siriuspy/siriuspy/cycle/conn.py
@@ -597,7 +597,7 @@ class LinacPSCycler:
 
     def cycle_duration(self, mode):
         """Return the duration of the cycling in seconds."""
-        _ = mode  # throwaway inherited argument
+        _ = mode  # throwaway arguments
         if self._cycle_duration is None:
             self._get_duration_and_waveform()
         return self._cycle_duration
@@ -622,7 +622,7 @@ class LinacPSCycler:
 
     def prepare(self, mode):
         """Config power supply to cycling mode."""
-        _ = mode  # throwaway inherited argument
+        _ = mode  # throwaway arguments
         status = True
         if not self.check_current_zero(wait=0.5):
             status &= self.set_current_zero()
@@ -630,7 +630,7 @@ class LinacPSCycler:
 
     def is_prepared(self, mode):
         """Return wether power supply is ready."""
-        _ = mode  # throwaway inherited argument
+        _ = mode  # throwaway arguments
         status = self.check_current_zero()
         return status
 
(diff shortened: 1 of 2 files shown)


siriuspy/siriuspy/cycle/conn.py
```diff
@@ -622,15 +622,15 @@ def check_current_zero(self, wait=5):
...
+        _ = mode  # throwaway arguments
         status = True
         if not self.check_current_zero(wait=0.5):
             status &= self.set_current_zero()
         return status
 
     def is_prepared(self, mode):
         """Return wether power supply is ready."""
-        _ = mode  # throwaway inherited argument
+        _ = mode  # throwaway arguments
```
fernandohds564: Why not just define the method this way:

`def is_prepared(self, _):`

xresende: > Why not just define the method this way:
> 
> `def is_prepared(self, _):`

much simpler. but you loose in-place information of the base class method arguments, in the case you want to extend the overridden method and make use of the inherited arguments. Naturally you can look up the base class API in this case.

You think we should standardize to your simple suggestion?

fernandohds564: I didn't understand this comment:
> much simpler. but you loose in-place information of the base class method arguments, in the case you want to extend the overridden method and make use of the inherited arguments. Naturally you can look up the base class API in this case.

Is this a base class for another one which will use the argument mode? If this is so, won't this method be implemented in the derived class again? The argument could be inserted then, don't you think?.
Or is this class derived from another one which has this argument? If this is the case, why does it keep the argument in its implementation? It could be completely removed now, couldn't it? Which use is being made of the inherited arguments?

> You think we should standardize to your simple suggestion?

I think if we really want to keep this argument, it is better to use this simpler version. Suppose you have your code folded in the class method level (generally level 2). This syntax allows a fast identification that the input variable does nothing in the method implementation.

xresende: > Is this a base class for another one which will use the argument mode? If this is so, won't this method be implemented in the derived class again? The argument could be inserted then, don't you think?.
> Or is this class derived from another one which has this argument? If this is the case, why does it keep the argument in its implementation? It could be completely removed now, couldn't it? Which use is being made of the inherited arguments?

I am considering it to be a derived class (subclass). what I meant is that it helps to have inherited argument names explicit in the method in case it needs to be improved/changed. But of course, we can always look at the method argument in the base class API if these method argument names are needed.

Without further much ado I will adopt the simpler version you proposed as our standard.

PR #430, opened 2020-03-30
Fix posang, opticscorr and currinfo tests

anacso17:


siriuspy/siriuspy/opticscorr/utils.py
```diff
@@ -1,43 +1,47 @@
...
+        self.acc = acc.lower()
+        self.opticsparam = opticsparam
+        self.fpath = '/home/sirius/iocs-log/' + self.acc + '-ap-' + \
+            opticsparam+'corr/'
+        self.fname = self.fpath + self.acc + '-' + opticsparam + 'corr.txt'
+
+    def get_config_name(self):
+        """Get the configuration name in file."""
+        try:
+            fil = open(self.fname, 'r')
```
fernandohds564: try to use context manager here. Instead of 

```
fil = open(self.fname, 'r')
config_name = fil.read().strip('\n')
fil.close()

```
you could simply do:
```
with  open(self.fname, 'r') as fil:
    config_name = fil.read().strip('\n')
```

siriuspy/siriuspy/opticscorr/utils.py
```diff
@@ -1,43 +1,47 @@
...
+                "'opticsparam' must be 'tune' or 'chrom' instance.")
+        self.acc = acc.lower()
+        self.opticsparam = opticsparam
+        self.fpath = '/home/sirius/iocs-log/' + self.acc + '-ap-' + \
+            opticsparam+'corr/'
+        self.fname = self.fpath + self.acc + '-' + opticsparam + 'corr.txt'
+
+    def get_config_name(self):
+        """Get the configuration name in file."""
+        try:
```
fernandohds564: you need this try..except just to check if the file exists? If this is the case, use:

```
import os
os.path.isfile(self.fname)
```
instead.

siriuspy/siriuspy/opticscorr/utils.py
```diff
@@ -1,43 +1,47 @@
...
+            fil = open(self.fname, 'r')
+            config_name = fil.read().strip('\n')
+            fil.close()
+        except Exception:
```
fernandohds564: I think it is better to be more specific on the `Exception` here...

siriuspy/siriuspy/posang/utils.py
```diff
@@ -1,28 +1,38 @@
...
+        self.fname = self.fpath + self.tl + '-posang.txt'
 
-def set_config_name(tl, config_name):
-    f = open('/home/sirius/iocs-log/' + tl + '-ap-posang/' +
-             tl + '-posang.txt', 'w+')
-    f.write(config_name)
-    f.close()
+    def get_config_name(self):
+        """Get config name from file."""
+        try:
```
fernandohds564: same comments here.

siriuspy/siriuspy/opticscorr/utils.py
anacso17: done!

siriuspy/siriuspy/opticscorr/utils.py
anacso17: done!

siriuspy/siriuspy/opticscorr/utils.py
anacso17: I removed the try..except!

PR #438, opened 2020-04-08
Excitation Data Rescaling Factor

xresende:
This PR allows for rescaling excitation curves for each power supply type. A configurable additional parameter `rescaling_factor` is added to each excdata file header. Its default value is 1.0

### DIFF
--- a/siriuspy/siriuspy/magnet/excdata.py
+++ b/siriuspy/siriuspy/magnet/excdata.py
@@ -171,11 +171,11 @@ class ExcitationData:
             xvals, xtab, ytab, left=float('nan'), right=float('inf'))
         nan = _np.isnan(interp)
         inf = _np.isinf(interp)
-        vec = _util.linear_extrapolation(xvals[nan],
-                                       xtab[0], xtab[1], ytab[0], ytab[1])
+        vec = _util.linear_extrapolation(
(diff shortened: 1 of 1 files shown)


siriuspy/siriuspy/magnet/excdata.py
```diff
@@ -165,37 +166,39 @@ def __str__(self):
...
+            xvals, xtab, ytab, left=float('nan'), right=float('inf'))
         nan = _np.isnan(interp)
         inf = _np.isinf(interp)
-        v = _util.linear_extrapolation(x[nan],
-                                       xt[0], xt[1], yt[0], yt[1])
-        interp[nan] = v
-        v = _util.linear_extrapolation(x[inf],
-                                       xt[-1], xt[-2], yt[-1], yt[-2])
-        interp[inf] = v
+        vec = _util.linear_extrapolation(xvals[nan],
```
fernandohds564, line 174: doesn't pylint complain about this?
either way, I think it is better to put this in the standard style, like line 170 and 171 above.

xresende: yes, pylint complains! the code was collapsed in the editor and I missed it. thanks.
fixed it, @fernandohds564.

PR #445, opened 2020-04-13
Improve SOFB Device

fernandohds564:
- Improve `devices.SOFB` to correctly deal with SinglePass and MultiTurn trajectories.

### DIFF
--- a/siriuspy/siriuspy/devices/device.py
+++ b/siriuspy/siriuspy/devices/device.py
@@ -91,14 +91,15 @@ class Device:
 
     def __getitem__(self, propty):
         """Return value of property."""
-        pvobj = self._pvs[propty]
-        value = pvobj.get()
-        return value
+        pvobj = self._pvs.get(propty)
+        if pvobj is not None:
+            return pvobj.get()
 
     def __setitem__(self, propty, value):
         """Set value of property."""
-        pvobj = self._pvs[propty]
-        pvobj.value = value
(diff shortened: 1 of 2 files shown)


siriuspy/siriuspy/devices/device.py
```diff
@@ -91,14 +91,15 @@ def values(self):
 
     def __getitem__(self, propty):
         """Return value of property."""
-        pvobj = self._pvs[propty]
-        value = pvobj.get()
-        return value
+        pvobj = self._pvs.get(propty)
```
xresende, line 94: pyling `inconsistent-return-statement`

siriuspy/siriuspy/devices/device.py
```diff
@@ -91,14 +91,15 @@ def values(self):
...
+        if pvobj is not None:
+            return pvobj.get()
 
     def __setitem__(self, propty, value):
```
xresende, line 98: same here

siriuspy/siriuspy/devices/device.py
fernandohds564: I don't understand. I don't get these errors.
What is it complaining about?

fernandohds564: done!

siriuspy/siriuspy/devices/device.py
fernandohds564: done!

PR #454, opened 2020-04-17
opticscorr.FIX: initialize Status and RefKL pvs correctly

anacso17:


anacso17:
@xresende yes, after verifying the other bugs!

PR #455, opened 2020-04-17
Fix bugs in OpticsCorr IOCs

anacso17:


siriuspy/siriuspy/opticscorr/chrom.py
```diff
@@ -389,7 +389,7 @@ def _callback_estimate_chrom(self, pvname, value, **kws):
         for fam in self._psfams:
             fam_idx = self._psfams.index(fam)
```
murilobalves, line 390: maybe this can be replaced by ``for fam_idx, fam in enumerate(self._psfams):``, what do you think?

anacso17: it seems to be clearer, I'll replace that!

PR #456, opened 2020-04-18
Add PV 'RF-Gen:GeneralFreq-SP' to configdb type global_config

xresende:


fernandohds564:
@xresende, could you also add the PV `RF-Gen:FreqPhsCont-Sel` to the `global_config`?
This PV is important to guarantee that phase will change continuously when we change the RF frequency, not causing problems to the LLRF control loops.
Is it correct, @felipekojihoshino?

felipekojihoshino:
Yes, It is correct. Please, add this one as well. Thanks!

xresende:
> @xresende, could you also add the PV `RF-Gen:FreqPhsCont-Sel` to the `global_config`?
> This PV is important to guarantee that phase will change continuously when we change the RF frequency, not causing problems to the LLRF control loops.
> Is it correct, @felipekojihoshino?

done.

PR #458, opened 2020-04-20
Pr devices syncd improve

xresende:


xresende:
> The only problem with `pvobj.value` is related to PVs without monitoring ( `-Mon`).

yes, that is true. I am not used to working without PV monitoring... Is really there a significative gain in not monitoring PVs like `_SP` and `_RB` ?! Whenever they change they should eventually be read in most application uses, anyway! Are there clear cases where monitoring by default lead to performance degradation or system overload?

PR #463, opened 2020-05-01
This PR adds and remove some BO and SI DLLRF PVs.

felipekojihoshino:


siriuspy/siriuspy/clientconfigdb/types/global_config.py
```diff
@@ -1231,6 +1229,33 @@ def get_dict():
     ['BR-RF-DLLRF-01:FF:DEADBAND:S', 0.0, 0.0],
     ['BR-RF-DLLRF-01:FF:GAIN:CELL2:S', 0.0, 0.0],
     ['BR-RF-DLLRF-01:FF:GAIN:CELL4:S', 0.0, 0.0],
+    ['BR-RF-DLLRF-01:AUTOSTART:S', 0.0, 0.0],
+    ['BR-RF-DLLRF-01:COMMSTART:S', 0.0, 0.0],
```
fernandohds564, line 1233: @felipekojihoshino, does this PV trigger some action of the IOC, or it just defines a state?
I'm asking because it is not recommend to include action PVs in configurations.

felipekojihoshino: Autostart can trigger an action according to the Commstart state and some other RF parameters read by LLRF.  At first it was one of the reasons that we preferred to not include it. I tought it could be useful at some point to support the system startup depending on the loaded configuration. Given that, it could be left both off for now. How can I do that? Should I just remove these lines and make another commit? I'm not really familiar with github yet...

PR #470, opened 2020-05-12
Pr devices device add wait for connection

xresende:


siriuspy/siriuspy/clientarch/devices.py
```diff
@@ -6,243 +6,19 @@
...
-        'SI-20M2:DI-BPM',
-        'SI-20C1:DI-BPM-1',
-        'SI-20C1:DI-BPM-2',
-        'SI-20C2:DI-BPM',
-        'SI-20C3:DI-BPM-1',
-        'SI-20C3:DI-BPM-2',
-        'SI-20C4:DI-BPM',
-        'SI-01M1:DI-BPM',
-        )
+    BPMS_TB = _BPMSearch.get_names({'sec': 'TB'})
```
fernandohds564, line 15: @xresende, there is an issue here. This commands will be executed in import time. Which means an error will be raised if there is no connection with the Constants server.

xresende: This is indeed not acceptable. Thanks!

PR #472, opened 2020-05-13
Move EthBrigdeClient import to machine_application repo

xresende:
This is a companion PR of https://github.com/lnls-sirius/machine-applications/pull/205:

- Move `EthBridgeClient` import to _machine_applications_, keeping _siriuspy_ free of hardware communication protocols other than EPICS.
- Change `requirements.txt` versioning conditionals in order to delegate precise control to _ansible_ configuration.

### DIFF
--- a/siriuspy/siriuspy/pwrsupply/pructrl/pru.py
+++ b/siriuspy/siriuspy/pwrsupply/pructrl/pru.py
@@ -60,7 +60,7 @@ class PRUInterface:
 class PRU(PRUInterface):
     """Functions for the programmable real-time unit."""
 
-    def __init__(self, ethbridgeclient, bbbname=None, ip_address=None):
+    def __init__(self, ethbridgeclnt_class, bbbname=None, ip_address=None):
         """Init method."""
         # if ip address was not given, get it from bbbname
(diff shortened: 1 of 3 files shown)


siriuspy/siriuspy/pwrsupply/pructrl/pru.py
```diff
@@ -71,12 +60,8 @@ def _close(self):
 class PRU(PRUInterface):
     """Functions for the programmable real-time unit."""
 
-    def __init__(self, bbbname=None, ip_address=None):
+    def __init__(self, ethbridgeclient, bbbname=None, ip_address=None):
```
fernandohds564, line 63: since this is not an instance of class but the class itself, I think it would be more appropriate to change the name of this variable to something like this: `ethbriggeclt_class` or `comm_class`

xresende: I agree. I though about it but was afraid the name would get too long. But it is better long than misleading...

xresende: done.

PR #483, opened 2020-06-01
APU22 Phase->Kx conversion

xresende:
Tested.


siriuspy/siriuspy/devices/psconv.py
```diff
@@ -197,11 +205,18 @@ def _get_devices(devname, proptype):
 
     @staticmethod
     def _get_dev_if_dipole(devname):
-        if devname.dev in {'B', 'B1B2'}:
+        if devname.dis in {'B', 'B1B2'}:
```
anacso17, line 208: Isn't that a bug?

xresende: It is! a very serious one! wondering how this happened... strange. thanks!!!

xresende: fixed it!

PR #486, opened 2020-06-03
Add IDSearch

anacso17:
This PR also:
- fix a bug in pwrsupply.csdev.py
- fix comments in util.conv_splims_labels
- add missing reset buffer command in SI Lifetime IOC

### DIFF
--- a/siriuspy/siriuspy/search/id_search.py
+++ b/siriuspy/siriuspy/search/id_search.py
@@ -1,0 +1,40 @@
+"""Search module."""
+
+
+class IDSearch:
+    """ID Search Class."""
+
+    _beamline2subsection = {
+        'Carnauba': '06SB',
+        'Caterete': '07SP',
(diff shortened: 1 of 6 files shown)


anacso17:
@xresende I will change the dictionary to be _beamline_name <-> id_name_!

siriuspy/siriuspy/search/id_search.py
```diff
@@ -0,0 +1,45 @@
+"""Search module."""
+
+
+class IDSearch:
+    """ID Search Class."""
+
+    _beamline2idname = {
```
fernandohds564, line 7: @anacso17, I think this should be a static table in control-system-constants, don't you?

xresende: @fernandohds564 , I think it is a good idea. I'll merge this PR as it is, add a NOTE and move tha table to csconstants later.

PR #489, opened 2020-06-04
Add auto_mon option do Devices class init

xresende:
While running tests with ID feedforward classes I realized that _-Mon_ PVs, for which original _Devices_ class turns `auto_mon` to False by default, were taking at least 20 ms in `PV.get`method. Here I added the option of  `auto_mon` to the class init. If set to True, `-Mon` PVs will also be created with  `auto_mon` on.

@fernandohds564 , are you using `auto_mon=False` for PVs in SOFB ? Did you not notice performance degradation?

ps: I just realized that the delay of 20 ms might be due to a test client running across a VPN connection...

### DIFF
--- a/siriuspy/siriuspy/devices/device.py
+++ b/siriuspy/siriuspy/devices/device.py
@@ -21,10 +21,6 @@ class Device:
 
     def __init__(self, devname, properties, auto_mon=False):
         """."""
-        # TODO: It takes ~ 20ms to get an -Mon PV update when auto_mon=False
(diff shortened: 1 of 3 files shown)


fernandohds564:
@xresende, in fact a call to `PV.get()` is slower when `auto_monitor` is False.
In `SOFB` I only use `aut_monitor=False` for large array PVs. I've never done a detailed analysis to know if it's worth to operate this way for such PVs, but I think that for simple PVs it's not.

PR #499, opened 2020-06-14
Replace isclose

xresende:
Replace np.isclose and math.isclose with functions that do exact single-precision float comparisons.


fernandohds564:
@xresende what about this PR?

xresende:
> @xresende the main reason for using these comparisons is timing delays and power supply waveforms. Even if we fix the power supply waveform comparisons, I think the cycle, ramp and epics.properties subpackages/modules need to maintain flexibility for timing comparisons.

I find checking applied PV from config very strange if the known precision is not taken into account. I have PS waveform in mind but it applies to timing PVs as well. 

these tolerance choices have some degree of arbitrariness, even if they pragmatically work now, (in principle) they may not work after a subsystem upgrade, for example. What happened recently with orbit tolerances in SOFB is an example of what I mean (although in this particular case I do not know if we have another robust solution other than defining indeed a pragmatic tolerance value)

Can we guarantee applied PV values up to know system precision (when known) another way, then ?

xresende:
> @xresende what about this PR?

resuming discussions...

fernandohds564:
what if we use the default value of the config to know up to which decimal place we need to compare?

PR #503, opened 2020-06-16
Change how SOFB signals not to use corrs in kicks and fix bugs in pssofb

xresende:


siriuspy/siriuspy/devices/pssofb.py
```diff
@@ -251,7 +251,7 @@ def current(self):
     @current.setter
     def current(self, value):
         """Set Current-SP vector in SOFB order."""
-        value = _np.array(value)
+        value = _np.asarray(value)
```
xresende, line 254: @fernandohds564 , `asarray` passes a symbolic link if `value` is already a numpy object. This may corrupt the invoking context's array. I think this is dangerous and should not be done since the diff of exec time is small (367 ns vs 246 ns, for 280-size arrays). This is way I did not use `asarray` in the setter method yet.

fernandohds564: I changed to `asarray` because I noticed `value` was not being changed inside this method. I think we could continue using it and in the future, if the implementation changes such that `value` is changed inside the method we drop it and go back to using `array`, what do you think?

xresende: fair.

PR #510, opened 2020-07-01
Fix antenna selection for BPM position algebra (singlepass)

xresende:
Fix antenna selection for BPM position algebra (singlepass) (on behalf of fernando h. de sá)


fernandohds564:
@xresende, I will merge this one, ok?

PR #514, opened 2020-07-06
ENH Devices: SOFB and Device

murilobalves:
- Add `MTTime-Mon` PV to get timestamp from SOFB
- Add `pvs` property to avoid dealing with private attributes. The motivation was the need for adding callbacks to pvs.

### DIFF
--- a/siriuspy/siriuspy/devices/device.py
+++ b/siriuspy/siriuspy/devices/device.py
@@ -44,6 +44,11 @@ class Device:
         return sims
 
     @property
+    def pvs(self):
+        """Return device PVs."""
+        return self._pvs
+
+    @property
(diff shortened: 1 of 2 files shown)


murilobalves:
> the new `pvs` property and the existing method `pv_object` are somewhat redundant. I don't see the need for them both existing at the same time. maybe they should be merged.

what do you think about the solution from last commit?

xresende:
> > the new `pvs` property and the existing method `pv_object` are somewhat redundant. I don't see the need for them both existing at the same time. maybe they should be merged.
> 
> what do you think about the solution from last commit?

I think it is aalright. did you run a grep to check whether `pv_object` is not yet being used in one of our repos?

PR #523, opened 2020-07-15
CurrInfo device

murilobalves:


murilobalves:
> @murilobalves, If I'm not wrong the `connected` property of this Device will always return `False`, right?
> Since the `CurrInfo` from the several sections are so different from each other, I think it is better to implement several classes, one for each section (maybe LI, TB and TS could share the same class), and then create a factory method or class to select between them depending on the Input.

I implemented the suggested changes, please take a look.

siriuspy/siriuspy/devices/currinfo.py
```diff
@@ -1,158 +1,222 @@
...
+        """."""
+        # check if device exists
+        if devname not in CurrInfoLinear.DEVICES.ALL:
+            raise NotImplementedError(devname)
+
+        # call base class constructor
+        super().__init__(devname, properties=CurrInfoLinear._properties)
+
+    @property
+    def transpeff(self):
```
fernandohds564, line 29: I think you could add the charge readings from the ICTs. They are very useful.

murilobalves: For LI, the ICTs readings are in ict device already. I could include the transport lines ICTs readings in ict device or merge the two devices in this currinfo (and add TB and TS ICTs readings). The transport efficiency are duplicated (in the currinfo IOC and in the ict IOC). I think it is better to use only the PVs provided by the currinfo IOC in the currinfo device and to add ICTs readings from TB and TS in the ict device (therefore related to icts IOCs only). What is your opinion?

fernandohds564: I think it would be useful to have the charges here too.. this way we don't have to create two devices everytime. You could add the ICT devices as subdevices of this one... But if you still feel uncomfortable with this option, no problem, you can leave it as is!

murilobalves: @fernandohds564 please take I look in the two last commits.

PR #538, opened 2020-07-21
Pssofb multiprocessing

fernandohds564:
tested!!

### DIFF
--- a/siriuspy/siriuspy/pwrsupply/pssofb.py
+++ b/siriuspy/siriuspy/pwrsupply/pssofb.py
@@ -681,7 +681,7 @@ class PSSOFB:
             bbbnames = PSSOFB.BBBNAMES[sub[i]:sub[i+1]]
             evt = _Event()
             evt.set()
-            mine, theirs = _Pipe(duplex=False)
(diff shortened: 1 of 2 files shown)


siriuspy/siriuspy/pwrsupply/pssofb.py
```diff
@@ -609,3 +581,200 @@ def _create_pru_udc(ethbridgeclnt_class):
...
+    def shutdown(self):
+        """."""
+        for pipe in self._pipes:
+            pipe.send(None)
+            pipe.close()
+        for proc in self._procs:
+            proc.join()
+
+    @staticmethod
+    def run_process(ethbridgeclnt_class, bbbnames, pipe, doneevt, shared_arr):
```
xresende, line 753: @fernandohds564 , I think this should be a private method, right?

fernandohds564: yes, it would be better that way

fernandohds564: done!

PR #540, opened 2020-07-23
Improve print statement for PS function return errors

xresende:


siriuspy/siriuspy/pwrsupply/pssofb.py
```diff
@@ -380,6 +400,9 @@ def _bsmp_current_setpoint(self, bbbname, curr_sp):
         # --- bsmp communication ---
         udc.sofb_current_set(tuple(setpoint))
 
+        # update sofb_func_return
+        self._sofb_func_return[indcs_sofb] = udc.sofb_func_return_get()
```
fernandohds564, line 404: I noticed this method may return `None`. In which situations this might occur? Should I consider `np.nan` a scenario where some of the indics of `_sofb_func_return` is `np.nan`? If I should, what should I do?

xresende: UDC class is a genreal classe that is used for communication with all power supply models. This UDC method returns `None` only if it is invoked with a non-FBP power supply UDC, which should not happen in any correct situation. The method will never return `None` for SOFB.

fernandohds564: But looking at the code, I noticed that is val1 or val2 is none, than it will also return none, even for fbps. Is This situation impossible to happen then?

xresende: There is code copied from another method that is unnecessary and can be deleted. It was misguiding your interpretation of all possible fluxes. I cleaned it up. val1 and val2 are always diferent from None.

PR #555, opened 2020-08-03
Add SOFBUpdate option to PSSOFB

xresende:


siriuspy/siriuspy/pwrsupply/pssofb.py
```diff
@@ -404,6 +409,12 @@ def _bsmp_current_setpoint(self, bbbname, curr_sp):
             return
         self._sofb_current_readback_ref[indcs_sofb] = current[indcs_bsmp]
 
+        # send signal to IOC to update one power supply state
+        if self._sofb_update_iocs:
+            pvname = self._sofbupdate[bbbname]
+            pvobj = _epics.get_pv(pvname)
```
fernandohds564, line 415: Isn't this slower than keeping a reference to the PV object?

xresende: I chose to use `get_pv` so as to avoid having to resolve the context in which PSConnSOFB is being used (inside PSSOFB, with processes or not?). If it is invoked in a separate process, get_pv wil automatically create a new PV connection at first invocation. It takes ~17ms in the first invocation and ~4us in subsequent ones, as measured with timeit in jupyter.

fernandohds564: ok then!

xresende: @fernandohds564 , I create the PV objects at the `init` method now.

fernandohds564 (approved):

murilobalves (approved):

anacso17 (approved):

PR #590, opened 2020-08-27
Add explicit PV object del in check_pv_online

xresende:


fernandohds564:
Was this creating some problem?

xresende:
> @xresende, was this creating some problem?

I am not sure. But the problem we obsorved with BO dipole PS IOC the other day, 'Virtual circuit unresponsive', appears to come from epics client lib (pyepics). And the only thing related to epics client is thi function being used in our IOCs. In any case this PR introduces a more controlled way to discard the PV object.

PR #594, opened 2020-09-02
Update IOC for new BO QF Firmware

xresende:


siriuspy/siriuspy/pwrsupply/bsmp/constants.py
```diff
@@ -12,7 +12,7 @@
 # version of the BSMP implementation of power supplies that is compatible
 # with the current implemenation of this module.
 
-__version__ = 'FRWM - V1.31 V0.41u - 8/19/2020'
+__version__ = 'FRWM - V1.33 V0.41u - 31/09/2020'
```
murilobalves, line 15: is 31/09/2020 right?

murilobalves (approved):

fernandohds564 (approved):

siriuspy/siriuspy/pwrsupply/bsmp/constants.py
xresende: @gabrielbrunheira  changed the formatting in the spec (excel file). I justed copied it so as to mirror the primary source of documentation.

gabrielbrunheira: There was indeed a mistake...it should have been 8/31/2020..I will update this file and send by email.

siriuspy/siriuspy/pwrsupply/bsmp/constants.py
```diff
@@ -416,40 +416,39 @@ class ConstFAC_2S_DCDC(ConstPSBSMP):
...
-    V_I_DRIVER_2_IIB_2 = 65
-    V_TEMP_BOARD_IIB_2 = 66
-    V_RH_IIB_2 = 67
-    V_IIB_INTERLOCKS_2 = 68
-    V_IIB_ALARMS_2 = 69
+    V_V_CAPBANK_1 = 36
+    V_V_CAPBANK_2 = 37
+    V_DUTY_CYCLE_1 = 38
+    V_DUTY_CYCLE_2 = 39
+    V_DUTY_DIFF = 40
```
gabrielbrunheira, line 423: This V_DUTY_DIFF doesn't exist anymore in the last BSMP specification

siriuspy/siriuspy/pwrsupply/bsmp/constants.py
gabrielbrunheira: I suggest use the newest version, **v1.36 V0.41 04/09/2020**

gabrielbrunheira (changes requested):

xresende:
@gabrielbrunheira , I am still confused with the versioning you suggest. in the latest excel file of the PS bsmp spec it is 'V0.41u'. up here you suggest 'V0.41'. which one should we use? also what is the date formatting we should adopt once-and-for-all? 09/04/2020 or 04/09/2020. since the excel formats dates according to users locale (english, portuguese, etc), this becomes ambiguous. 

Naturally it does not matter the least which format we use but we should make a choice and keep it, so that issues like that do not take our time anymore. I suggest we keep english formatting, 09/04/2020. what do you think?

PR #660, opened 2021-02-28
Generalize SOFBMode

xresende:
To allow remote power supply firmware updates a **StandBy mode for the IOCs was resquested**.
This mode is exactly the same as the SOFBMode=On mode already implemented to FBP PSs.
The idea here is to reuse this SOFBMode mechanism by generalizing it to all types of power supplies.

1. This is an initial version of the PR. **It just renames  SOFBMode to StandByMode**
2. The PV was added for all PS types epics dbases.
3. SOFMode PV name as kept for SOFB classes
4. We will need to rename PVs in archiver
5. There are  PRs for HLA and Machinne-applications to be done.

ps1: @anacso17 , @fernandohds564 , we can discuss this PR along the week.
ps2: Changes from PR https://github.com/lnls-sirius/dev-packages/pull/659 were already committed into this PR, since they deal with 'SOFMode' names, and would be affected by renaming. Please ignore these changes for the purpose of mode renaming...


fernandohds564:
@xresende, is this still a thing? If it does, what is missing to be ready for review?

anacso17:
This seems to be deprecated, considering https://github.com/lnls-sirius/dev-packages/pull/1108

PR #662, opened 2021-03-02
Fix SOFB

fernandohds564:
This PR tries to solve the problem with SOFB initialization, where some of the child processes, sometimes, do nothing after their creation.

I also removed some imports from the `timesys` subpackage to remedy cyclic import problems.

Maybe we should move the `csdev` modules to a separated subpackage, just as it was in the past.


siriuspy/siriuspy/sofb/orbit.py
```diff
@@ -22,6 +22,23 @@
...
+    """
+    def __init__(self, **kws):
+        _mp.Process.__init__(self, **kws)
+
+    def run(self):
+        _ca.initial_context = None
+        _ca.clear_cache()
+        _mp.Process.run(self)
+
+
```
anacso17, line 41: It would be interesting to define a kind of "ProcessSpawn" class in the siriuspy.epics subpackage to avoid repeated derivations, wouldn't it?

fernandohds564: nice @anacso17 ! I'll do it"

fernandohds564: done!

anacso17 (approved):

xresende (approved):

murilobalves (approved):

PR #686, opened 2021-04-15
Improve clientarch subpackage

anacso17:
Summary: 
- improve Time class
  - subclass Time from datetime
  - add multiple constructor options 
  - make \_\_add\_\_ and \_\_sub\_\_ magic methods to return Time objects and to handle operations with float and int types
  - add auxiliary function get_time_intervals for parallelization

-  add new version of ClientArchiver
   - use aiohttp and asyncio in requests for better performance
   - add option to pass a list/tuple of timestamps to perform parallelized requests
   - implement connected method
   - add logout method to properly close the connection when the user logs in

- add aiohttp to requirements

- improve PVData class
   - handle timestamps with Time class
   - add properties to handle attributes
   - add time_start and time_stop properties
   - handle parallelization intervals in update method

- add new class PVDataSet


murilobalves (approved):

requirements.txt
```diff
@@ -4,3 +4,4 @@ pyepics>=3.4.0
 python-dateutil>=2.8.1
 requests>=2.22.0
 six>=1.13.0
+aiohttp>=3.7.4.post0
```
xresende, line 7: suggestion: drop bugfix subversioning

xresende (approved):

requirements.txt
anacso17: done!

siriuspy/siriuspy/clientarch/time.py
```diff
@@ -1,84 +1,51 @@
...
+            add = super().__add__(_timedelta(seconds=other))
+        else:
+            add = super().__add__(other)
+        return Time(timestamp=add.timestamp())
 
     def __sub__(self, other):
         """Subtraction."""
-        return self.datetime - other
+        if isinstance(other, _datetime):
+            sub = super().__sub__(other)
```
fernandohds564, line 45: there is no need to create a temporary variable here.
`return super().__sub__(other)`

siriuspy/siriuspy/clientarch/time.py
```diff
@@ -1,84 +1,51 @@
...
     def get_iso8601(self):
         """Get iso8601 format."""
-        return self.datetime.astimezone().isoformat()
+        return self.astimezone().isoformat()
 
     def __add__(self, other):
         """Addition."""
-        return self.datetime + other
+        if isinstance(other, _datetime):
+            add = super().__add__(other)
```
fernandohds564, line 34: there is no need to create a temporary variable here.
return super().__add__(other)

siriuspy/siriuspy/clientarch/time.py
```diff
@@ -1,84 +1,51 @@
...
-            if not isinstance(datetime, _datetime):
-                raise TypeError("Expected an argument of type 'datetime'.")
-            self.datetime = datetime
-        elif timestamp:
+    def __new__(cls, year=None, month=None, day=None,
+                hour=None, minute=0, second=0, microsecond=0,
+                timestamp=None,
+                timestamp_string=None, timestamp_format=None):
+        """New."""
+        if timestamp:
```
fernandohds564, line 14: I think `timestamp` is the name of the string, I think it is confusing to use it as a `float` here. Maybe the name of the `float` argument could be `time`?

Also, wouldn't it be useful if the `time` argument were the first keyword argument, such that it would be possible to call:
`Time(time.time())`

What happens if we call `Time()`? Is this equivalent to `Time.now()`?

siriuspy/siriuspy/clientarch/pvarch.py
```diff
@@ -113,55 +114,155 @@ def __str__(self):
...
+
+    @property
+    def timestamp_stop(self):
+        """Timestamp stop."""
+        if not self._time_stop:
+            return None
+        return self._time_stop.timestamp()
+
+    @timestamp_stop.setter
+    def timestamp_stop(self, new_timestamp):
```
fernandohds564, line 202: Same here, I think `timestamp_(start|stop)` should be strings. You could accept float or integers in `time_(start|stop)`.

fernandohds564 (changes requested):
@anacso17  I started my review from end to start. I think if you follow the comments in order of appearance they may sound strange :smile:

siriuspy/siriuspy/clientarch/time.py
anacso17: > What happens if we call `Time()`? Is this equivalent to `Time.now()`?

No, in the current implementation this sentence returns an error, as like the super class. Do you think it is useful/intuitive this implementation?

siriuspy/siriuspy/clientarch/time.py
anacso17: done!

siriuspy/siriuspy/clientarch/time.py
anacso17: done!

fernandohds564 (approved):

siriuspy/siriuspy/clientarch/time.py
anacso17: (offline talk summary)
We decided to keep `timestamp` name for float inputs and `timestamp_string` for string inputs to maintain consistency with the superclass nomenclature

siriuspy/siriuspy/clientarch/pvarch.py
anacso17: idem of previous comment.

siriuspy/siriuspy/clientarch/time.py
anacso17: > Also, wouldn't it be useful if the time argument were the first keyword argument, such that it would be possible to call:
Time(time.time())

done!

fernandohds564 (approved):

PR #701, opened 2021-06-16
Derive PRUInterface from IOInterface

xresende:


siriuspy/siriuspy/pwrsupply/bsmp/commands.py
```diff
@@ -58,14 +63,15 @@ class PSBSMP(_BSMP):
             CONST.V_WFMREF1_IDX),
     }
 
-    def __init__(self, slave_address, entities, pru=None):
+    def __init__(self, slave_address: int, entities: _EntitiesPS, pru: _IOInterface):
         """Init BSMP."""
-        self.pru = pru
+        self.pru: _IOInterface = pru
+        self.entities: _EntitiesPS
```
xresende, line 68: is this annotation necessary ?

carneirofc: It is because we are accessing the member `parameters` which is defined at the child class `EntitiesPS` intead of the base `bsmp.Entities`.

Here for example https://github.com/lnls-sirius/dev-packages/pull/701/files#diff-6852f94bce2009b4831beae5cdc36c289191acda8bf2de952278dd167cbcd9f9R192 the linter would assume that entities is of type `bsmp.Entities` instead of `pwrsupply.bsmp.EntitiesPS`.

This is not the ideal solution but will fix eventual static analyser errors ([mypy](https://mypy.readthedocs.io/en/stable/) ).

carneirofc (approved):

fernandohds564 (approved):

anacso17 (approved):

murilobalves (approved):

carneirofc:
https://github.com/lnls-sirius/dev-packages/blob/dd91bb2fa94f43a8240f3615c27fa376cd4f0d30/siriuspy/siriuspy/pwrsupply/pructrl/pru.py#L21

This is probably an error, we should add  
```python
pass
```

PR #709, opened 2021-07-02
Improve devices

anacso17:
* Add first version of Injection System Standby Handlers to siriuspy.devices


xresende (approved):

murilobalves (approved):

siriuspy/siriuspy/devices/lillrf.py
```diff
@@ -40,13 +42,31 @@ def amplitude(self, value):
...
+    def integral(self):
+        """Integral Enable."""
+        return self['GET_INTEGRAL_ENABLE']
+
+    @integral.setter
+    def integral(self, value):
+        self['SET_INTEGRAL_ENABLE'] = value
+
+    @property
+    def feedback(self):
```
fernandohds564, line 62: I think it is better to use `feedback_state`. What do you think?

siriuspy/siriuspy/devices/lillrf.py
```diff
@@ -40,13 +42,31 @@ def amplitude(self, value):
...
-        """."""
+        """Phase."""
         return self['GET_PHASE']
 
     @phase.setter
     def phase(self, value):
         self['SET_PHASE'] = value
 
+    @property
+    def integral(self):
```
fernandohds564, line 53: I think it is better to use `integral_enable`. What do you think?

siriuspy/siriuspy/devices/lillrf.py
```diff
@@ -40,13 +42,31 @@ def amplitude(self, value):
...
+    @property
+    def feedback(self):
+        """Feedback Mode."""
+        return self['GET_FB_MODE']
+
+    @feedback.setter
+    def feedback(self, value):
+        self['SET_FB_MODE'] = value
+
     def cmd_set_phase(self, value, timeout=10):
```
fernandohds564, line 70: I know this wasn't created by you, but I don't like this method. It is not a `cmd` maybe we coulde remove it and leave only the setter function. Maybe it would be useful to have commands to turn on and off the feedback and the integral. What do you think?

siriuspy/siriuspy/devices/timing.py
```diff
@@ -191,3 +193,120 @@ def cmd_external_trigger(self):
...
+
+    def __init__(self, trigname):
+        """Init."""
+        self._database = _get_hl_trigger_database(trigname)
+        self._properties = tuple(self._database)
+        self._source_options = self._database['Src-Sel']['enums']
+        super().__init__(trigname, properties=self._properties)
+
+    @property
+    def status(self):
```
fernandohds564, line 212: maybe we could add a property with the description of the status? for example, if the status is ok, it would return a string "Ok", if bit 1 and 3 indicates problem, it would return a string with the description of each bit, in a comma separated form. What do you think?

fernandohds564 (approved):

siriuspy/siriuspy/devices/lillrf.py
anacso17: I agree!

siriuspy/siriuspy/devices/lillrf.py
anacso17: agree!

siriuspy/siriuspy/devices/lillrf.py
anacso17: do you think we can move the `wait` here to the property setter?

anacso17: > Maybe it would be useful to have commands to turn on and off the feedback and the integral. 

one command for each property? the command would wait for loop to be "well closed"?

siriuspy/siriuspy/devices/timing.py
anacso17: good idea, working on that

siriuspy/siriuspy/devices/lillrf.py
fernandohds564: > do you think we can move the `wait` here to the property setter?

humm, I don't remember how we do it on other classes... maybe we could copy.

> one command for each property? 

yes

>the command would wait for loop to be "well closed"?

no, I don't think it is desirable for it to wait. But it would be useful to have a method to wait for it and also to check if the loop is well closed within some chosen range.

siriuspy/siriuspy/devices/injsys.py
```diff
@@ -0,0 +1,623 @@
...
+
+        # pu devices
+        self._pudevs = [PowerSupplyPU(pun) for pun in self._punames]
+
+        # trigger devices
+        self._trigdevs = [Trigger(trg) for trg in self._trignames]
+
+        alldevs = list(self._pudevs)
+        alldevs.extend(list(self._trigdevs))
+        alldevs = tuple(alldevs)
```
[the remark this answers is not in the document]
anacso17: done!

siriuspy/siriuspy/devices/injsys.py
```diff
@@ -0,0 +1,623 @@
...
+
+        # ps devices
+        self._psdevs = [PowerSupply(psn) for psn in self._psnames]
+
+        # trigger devices
+        self._trigdevs = [Trigger(trg) for trg in self._trignames]
+
+        alldevs = list(self._psdevs)
+        alldevs.extend(self._trigdevs)
+        alldevs = tuple(alldevs)
```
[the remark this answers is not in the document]
anacso17: done!

siriuspy/siriuspy/devices/lillrf.py
anacso17: done!

siriuspy/siriuspy/devices/lillrf.py
anacso17: done!

siriuspy/siriuspy/devices/lillrf.py
anacso17: @fernandohds564 can you please take a look at my first version for these methods?

siriuspy/siriuspy/devices/timing.py
anacso17: done, please take a look!

siriuspy/siriuspy/devices/lillrf.py
```diff
@@ -40,25 +45,80 @@ def amplitude(self, value):
...
+        """Set and wait for amplitude property to reach value."""
         self.amplitude = value
         self._wait_rb_sp(timeout, 'amplitude')
 
+    def cmd_set_integral_enable(self, value, timeout=3):
+        """Set and wait for integral enable property to reach value."""
+        self.integral_enable = value
+        self._wait('GET_INTEGRAL_ENABLE', value, timeout=timeout)
+
+    def cmd_set_feedback_state(self, value, timeout=3):
```
fernandohds564, line 108: @anacso17 , my ideia was too have cmds to turn on and turn off the feedback and the integral enable.

siriuspy/siriuspy/devices/timing.py
fernandohds564: Very nice @anacso17 !

siriuspy/siriuspy/devices/lillrf.py
anacso17: oh, I see, together in the same method, just like injection system handlers?

fernandohds564: like the ones we have in other Devices.

xresende:
I just realized now that a PR in siriuspy changes the method names in RFCav device: from 'cmd_set_phase' to 'set_phase'. But other devices kept using the prefix 'cmd_'. I think we should try to keep consistent notation. I think prefixes are useful in order to,  in completion editing actions, group methods  that share characteristics. In this particular case, methods that evaluate to device commands. I would keep 'cmd_' but, if not, I think we should then rename the corresponding methods from other device objects.

xresende (changes requested):
I just realized now that a PR in siriuspy changes the method names in RFCav device: from 'cmd_set_phase' to 'set_phase'. But other devices kept using the prefix 'cmd_'. I think we should try to keep consistent notation. I think prefixes are useful in order to, in completion editing actions, group methods that share characteristics. In this particular case, methods that evaluate to device commands. I would keep 'cmd_' but, if not, I think we should then rename the corresponding methods from other device objects.

anacso17:
@xresende I think the idea of renaming is because these commands are actually setpoints (one argument/input is required). Following our Sirius PV naming convention, commands (-Cmd) do not require arguments, while setpoints (-SP, -Sel) do. So the idea of renaming is actually to use a more consistent notation.

xresende:
@anacso17 , originally when I created 'cmd_' methods in device classes ( I think in PowerSupply) I had no intention of them being equivalent to '-Cmd' PVs. The idea was that 'command' methods would gather device actions through various PVs, either '-Cmd'-type PVs or otherwise, and the prefix would help to completion-gather methods of the same class. There is no reason why a action method of a device should be attached to only one -Cmd PV, and hence I find this naming equivalence rather arbitrary. moreover, it is too restrictive to associate class methods to only one PV and it goes against idea of using Devices for interacting with the hardware, instead of using plain PVs.

But we are to drop 'cmd_' prefix (for whatever particular reasoning) I suggest we be consistent and drop it over all other devices as well. Over PowerSupply, for example.

anacso17:
@xresende what I meant was that the idea of the renaming was to keep methods prefixed with `cmd_` with an interface equivalent to PVs from -Cmd, not to just use these methods for PVs of the type -Cmd, nor for just one of them. The idea of the renaming is: if the command needs an argument/input, it is a setpoint, and could be grouped with `set_` prefixes, not `cmd_`. It's just a differentiation between the method API using the prefix.

fernandohds564:
I think the idea is what @anacso17 explained, @xresende . This change I made here was to keep things consistent, not to break an existing one. 

Follow below my answer to your comment in the apsuite PR:

@xresende , the cmd_ prefix is used for methods that do not take inputs (apart from timeout). They are commands that always execute the same task. I removed the cmd_ from these methods because the are setpoint methods and do take inputs.

xresende:
@anacso17 , @fernandohds564 , ok, so we keep 'cmd_' prefix only for methods without arguments. I guess the original naming of PowerSupply methods already complies with this convention.

xresende (approved):

PR #713, opened 2021-07-17
Fix PS IOCs timeout bug

xresende:
Intercept newly introduced EthBridge socket.timeout exception so at to avoid comm. operation thread in PS IOC to die.


anacso17 (approved):

fernandohds564 (approved):

siriuspy/siriuspy/pwrsupply/pructrl/prucontroller.py
```diff
@@ -5,8 +5,8 @@
 at the other end of the serial line.
 """
 
-import time as _time
-import socket as _socket
+from time import (time as _time, sleep as _sleep)
```
fernandohds564, line 8: I think you don't need the parenthesis here. We usually don't use it in this situation.

murilobalves (approved):

siriuspy/siriuspy/pwrsupply/pructrl/prucontroller.py
xresende: yes, they are not necessary. It is not this case, but I find them useful for splitting import lines with a better syntax (without the need of \ char) when the number of symbols is larger, just like we split statement lines with lists and tuples. I was reminded of its use by one of claudio's PR and just started using this approach. I also found using parenthesis more easy to read, since somehow they impinge the symmetry between imported symbols onto the import syntax.

by the way, the idea behind this last commit was to avoid importing the whole "time" module, to give the interpreter a chance to import only necessary symbols, if that can be optimized somehow by it.

fernandohds564: does it optimize runtime performance, or just the module loading time?

xresende: I think, if it does at all, just the loading time.

pnallin (approved):

PR #747, opened 2021-10-21
Add enable/disable PVs to Tune dev

xresende:


siriuspy/siriuspy/devices/tune.py
```diff
@@ -104,6 +123,31 @@ def tuney_wfm(self):
         """."""
         return self.devices[3].tune_wfm
 
+    @property
+    def enablex(self):
+        """."""
+        return self.devices[0]['Enbl-Sts']
```
fernandohds564, line 129: I think it is better to use the property enable here, don't you think?

siriuspy/siriuspy/devices/tune.py
```diff
@@ -104,6 +123,31 @@ def tuney_wfm(self):
...
+
+    @property
+    def enabley(self):
+        """."""
+        return self.devices[1]['Enbl-Sts']
```
fernandohds564, line 134: Same here.

PR #778, opened 2022-01-07
Add MeasFillingPattern to currinfo

xresende:
Add a rudimentary but working class for measurement of beam filling patterns (from andrei's script)

### DIFF
--- a/siriuspy/siriuspy/currinfo/filling_pattern.py
+++ b/siriuspy/siriuspy/currinfo/filling_pattern.py
@@ -17,8 +17,10 @@ class MeasFillingPattern:
         SI = ('10.128.150.77', 5025)
         ALL = (SI, )
 
-    def __init__(self, device):
+    def __init__(self, device=None):
         """."""
+        device = device or MeasFillingPattern.DEVICES.SI
+
         # check if device exists
         if device not in MeasFillingPattern.DEVICES.ALL:
             raise NotImplementedError(device)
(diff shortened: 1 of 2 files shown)


fernandohds564 (approved):

murilobalves (approved):

anacso17 (approved):

anacso17 (approved):

murilobalves (approved):

siriuspy/siriuspy/currinfo/filling_pattern.py
```diff
@@ -17,14 +17,14 @@ class DEVICES:
         SI = ('10.128.150.77', 5035)
         ALL = (SI, )
 
-    def __init__(self, devname):
+    def __init__(self, device):
```
fernandohds564, line 20: I think it is better to make this an optional argument.

xresende: you think so? the way I see the class, despite its name, is quite general. it could be used to get any keysight osc. waveform, right, as long as the IP, PORT tuple is passed, right ? you suggest using DEVICES.SI as default?

fernandohds564: Yes, I suggest using `DEVICES.SI` as default since it greatly simplifies the creation of instances. 
Apart from that, I don't think this class is general, because the methods defined here for interaction with the Oscilloscopes are particular for this application. I understand what you mean, though, considering how easy it would be to use this code to make other classes.

xresende: Yes, what I had in mind what eventually create a general class for Keysight oscilloscopes, without any default `device` argument, and rewrite `MeasFittingPattern` as a specialized implementation of this class. 
Ok, I have added DEVICES.SI as default device for the current `MeasFittingPattern` class.

PR #780, opened 2022-01-10
Add ClientArchiver timeout setter/getter

xresende:


murilobalves (approved):

fernandohds564 (commented):
@anacso17, @xresende what do you think about changing the interface so that an error is raised in case of timeout?

anacso17:
> @anacso17, @xresende what do you think about changing the interface so that an error is raised in case of timeout?

We were indeed talking about this problem of not having a well documented return in case of empty returns. I think raising an error is a good choice for timeout cases

xresende:
> @anacso17, @xresende what do you think about changing the interface so that an error is raised in case of timeout?

raising Timeout error.

fernandohds564 (approved):

anacso17 (approved):

fernandohds564 (approved):

anacso17:
Hi guys, I just added some commits that:
- Fixed a bug observed when closing a Qt application that uses ClientArchiver
- Make ClientArchiver compatible for waveform data
- Replace plot_date with xaxis.axis_date, because matplotlib will discontinue plot_date in the future
- Change PVData to simply return None in case ClientArchiver receives an empty response

murilobalves (approved):

PR #781, opened 2022-01-11
pwrsupply.ENH: add ParamUpdate-Cmd PV

anacso17:
This PR adds a new ParamUpdate-Cmd PV that triggers a new power supply parameter reading. Currently, this task is only performed on IOC initialization (when the corresponding Param*-Cte PVs are filled) because it takes ~0.7s per power supply. The idea of this new command is to make it possible to update these PVs "on the fly" during maintenance activities, when power supply parameters are updated, such as in the following weeks.


siriuspy/siriuspy/pwrsupply/psctrl/pscwriters.py
```diff
@@ -411,6 +411,19 @@ def execute(self, value=None):
             self.pru_controller.sofb_update_variables_state()
 
 
+class ParamUpdate(Function):
+    """ParamUpdate Function."""
+
+    def __init__(self, device_ids, pru_controller):
+        """Create command to update parameters."""
+        self.device_ids = device_ids
```
xresende, line 419: maybe use `self._devices_ids` instead. And change `class Voltage` that is not standardized either.

xresende (changes requested):

fernandohds564 (approved):

murilobalves (approved):

siriuspy/siriuspy/pwrsupply/pructrl/prucontroller.py
```diff
@@ -264,6 +264,21 @@ def exec_functions(self, device_ids, function_id, args=None):
         self._queue.append(operation)
         return True
 
+    def update_parameters(self, device_ids):
+        """Update device parameters."""
+        # if in SOFBMode on, do not accept comm. commands
+        if self._sofb_mode:
```
fernandohds564, line 270: the last commit got me confused about the motivation of this PV. I thought it was created to update the parameters of the Power Supply when it were in SOFB mode, because otherwise they would always be updated automatically. But it seems this is not the case.

anacso17: The parameters we are handling here are those related to "-Cte" PVs that we show in the PS Parameters auxiliary window. The reading of those parameters is an expensive task (it takes ~0.7s per power supply) and is only performed in IOC initialization. The idea here is to make it possible to update these PVs "on the fly" during maintenance activities, such as in the following weeks.

anacso17: Sorry guys for not adding a description for this PR

siriuspy/siriuspy/pwrsupply/pructrl/prucontroller.py
```diff
@@ -795,3 +807,10 @@ def _dict2list_vargroups(groups_dict):
         # create list of variable ids
         groups_list = [groups_dict[gid] for gid in group_ids]
         return groups_list
+
+    def _read_parameter_values(self, device_ids=None):
```
xresende, line 811: @anacso17 , I think you should change the method name to `_bsmp_read_parameter_values`. All primary class methods that do bsmp comunication have names starting with prefix `_bsmp`. See

https://github.com/lnls-sirius/dev-packages/blob/011c4f6d471b0c0ada3b24d4e7804b58bf1dd622/siriuspy/siriuspy/pwrsupply/pructrl/prucontroller.py#L32

xresende (changes requested):

siriuspy/siriuspy/pwrsupply/pructrl/prucontroller.py
anacso17: Really forgot that detail, fixed! I also changed the order of the methods to group this one with the other `_bsmp` methods

xresende (approved):

PR #811, opened 2022-05-12
Add fast corrector controls

anacso17:
- Add fast correctors to cycle procedure
- Add PSDiag PVs for fast correctors
- Add PSConv data for fast correctors (pv database)
- Add si_fofb config type to clientconfigdb


siriuspy/siriuspy/cycle/fc_cycle_data.py
```diff
@@ -0,0 +1,187 @@
+"""SI fast correctors cycle data."""
+
+import numpy as _np
+
+
+PARMS = {
```
fernandohds564, line 6: @anacso17 why did we choose to put this table here and not on CS-constants?

fernandohds564 (approved):

siriuspy/siriuspy/cycle/conn.py
```diff
@@ -753,3 +754,148 @@ def _get_duration_and_waveform(self):
...
+        """Config power supply to cycling mode."""
+        status = True
+        if not self.check_current_zero(wait=0.5):
+            status &= self.set_current_zero()
+        return status
+
+    def is_prepared(self, _, wait=5):
+        """Return whether power supply is ready."""
+        status = self.check_current_zero(wait)
+        return status
```
murilobalves, line 851: in this case you can do `return self.check_current_zero(wait)` without the need of `status` variable, right?

murilobalves (approved):

siriuspy/siriuspy/cycle/conn.py
anacso17: right, done!

siriuspy/siriuspy/cycle/fc_cycle_data.py
fernandohds564: I don't have a strong opinion about it. I just asked because I thought all parameters for cycling were at the control-system-constants repository. Since Linac parameters are already here, maybe we could leave these ones here too...

murilobalves (approved):

siriuspy/siriuspy/cycle/fc_cycle_data.py
xresende: I would also expect it to be in control-system-constants. maybe we have move it ove rthere later...

xresende (approved):

PR #817, opened 2022-06-03
Fix InjCtrl IOC

anacso17:
This PR fixes a bug noticed during last week's operation. Now, Injection Control IOC does not allow multiple threads to run the same EGun settings at the same time.

### DIFF
--- a/siriuspy/siriuspy/injctrl/main.py
+++ b/siriuspy/siriuspy/injctrl/main.py
@@ -291,6 +291,12 @@ class App(_Callback):
             self._update_log(
                 'ERR:Turn off top-up mode before changing inj.type.')
             return False
+        if self._thread_type is not None and self._thread_type.is_alive():
+            self._update_log('ERR:New setpoint received for injection')
+            self._update_log('ERR:type, but still processing the')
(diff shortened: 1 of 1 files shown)


fernandohds564 (changes requested):
I don't think this is the desired behavior. I think the IOC should interrupt the last command and start executing the new one. It is very common for us to set wrong values and try to correct things before the ramp finishes. The way you implemented here, the EGun would first reach the final, possibly wrong destination and only then would go back.

anacso17:
> I don't think this is the desired behavior. I think the IOC should interrupt the last command and start executing the new one. It is very common for us to set wrong values and try to correct things before the ramp finishes. The way you implemented here, the EGun would first reach the final, possibly wrong destination and only then would go back.

Thanks @fernandohds564 , I will implement this way!

murilobalves (approved):

anacso17:
Thanks @fernandohds564 for your suggestions!

xresende:
@anacso17 , did you take the time to analyze what PVs of this IOC are useful to be added to the epics archiver ?

xresende (approved):

PR #822, opened 2022-06-30
Allow removal of correctors in PSSOFB mode

fernandohds564:
# Changes made:
 - PS -> Improve interpolation in excdata:
    - Only call linear_extrapolation when needed.
   - When xvals is NaN interp will be NaN: before this change the code wasn't able to separate NaN inputs from left extrapolation.
- PSSOFB -> Only send signal to IOC to update variables when PV is connected.
- PSBSMP -> Handle NaNs in setpoints properly:
   - If values to be set has NaNs, filter them;
   - If all entries of values are NaNs, do not send setpoints and return ok.
- SOFB -> When corrector is not being used in PSSOFB mode, try not to send setpoints to it.
- SOFB -> Do not check status or configure correctors and BPMs not used in correction.

# Tests needed before merging this PR:

 - Launch IOC without errros;
 - Configure BPM and correctors parameters and check if parameters of correctors and BPMs not used were not changed;
 - Change the some parameter of BPM and corrector not in loop and check if status is still ok;
 - Close the loop with all BPMs and Correctors selected in PSSOFB mode;
 - Open the loop, remove a corrector and close the loop again;
 - Remove a corrector without opening the loop;
 - Disconnect some UDC from a beaglebone and try to close the loop;
 - Disconnect some Beaglebone from ethernet and try to close the loop;


xresende (approved):

siriuspy/siriuspy/pwrsupply/bsmp/commands.py
```diff
@@ -574,7 +574,8 @@ def sofb_ps_func_return(self):
     def sofb_ps_setpoint_set(self, value):
         """."""
         if self._sofb_ps_setpoint is None:
-            self._sofb_ps_setpoint = _np.zeros(value.size, dtype=float)
```
xresende: @fernandohds564 , this const has always the same value, 4. This is just the maximum number of BSMP devices that each UDC that we have today can control. This value is 4 even if UDC controls 3 or less devices.

fernandohds564: I know @xresende, that's why we always send setpoints with arrays of size four.

xresende (approved):

xresende:
@fernandohds564 , @anacso17 

I add excitation curve interpolation of power supplies other than SI correctors to the list of tests needed.

anacso17 (approved):

PR #823, opened 2022-07-01
Add oscilloscope

xresende:
Implementation of a temporary class to be used for oscilloscope acquisitions. Andrei is looking for python packages to control scopes, in particular with Keysight scopes in the beginning.


fernandohds564 (approved):

siriuspy/siriuspy/oscilloscope/keysight.py
```diff
@@ -0,0 +1,203 @@
...
+        return wavet, waved
+
+    def stat_init(self):
+        """Enable scope measurement statistics info."""
+        sock = self._socket
+        # Set bit order to MSB First
+        sock.sendall(b":MEASure:STATistics ON\n")
+        sock.sendall(b":MEASure:SENDvalid ON\n")
+
+    def stat_acquire(self):
```
murilobalves, line 170: at a first sight I thought that there was a typo here and it should be "start_acquire", then I saw the docstring explaining that `stat` is a short for "statistics". Maybe rename this methods to avoid confusion? `acquire_stats` or `stats_acquire`. Also the other method to: `init_stats`, `stats_init`, for this case I would suggest something like `enable_stats` following the docstring description.

murilobalves (approved):

siriuspy/siriuspy/oscilloscope/keysight.py
xresende: "we'll do"

murilobalves (approved):

PR #826, opened 2022-07-22
Improve devices and add new PV to control PU configuration in InjCtrl IOC

anacso17:
Hi guys, this PR:
- Adds first version of PosAng device
- Adds first version of InjSysPUModeHandler, that handles the pulsed magnets and PosAng to switch between Accumulation, Optimization and OnAxis configurations
- Adds new PUMode-Sel/Sts PVs in InjCtrl IOC to control InjSysPUModeHandler
- Cleanup code and improve docstrings

### DIFF
--- a/siriuspy/siriuspy/devices/injsys.py
+++ b/siriuspy/siriuspy/devices/injsys.py
@@ -807,6 +807,8 @@ class InjSysPUModeHandler(_Devices, _Callback):
     @property
     def is_accum(self):
         """If configuration is Accumulation."""
+        if not self.connected:
+            return False
         dpk_off = (self.pudpk.pwrstate == PowerSupplyPU.PWRSTATE.Off) or \
             (self.pudpk.pulse == PowerSupplyPU.PULSTATE.Off)
         nlk_on = (self.punlk.pwrstate == PowerSupplyPU.PWRSTATE.On) and \
@@ -845,6 +847,8 @@ class InjSysPUModeHandler(_Devices, _Callback):
     @property
     def is_optim(self):
         """If configuration is Optimization."""
+        if not self.connected:
+            return False
         dpk_evt_ok = self.trigdpk.source == self.trignlk.source
         dpk_dly_ok = self.trigdpk.delay_raw == self.SI_DPKCKR_DLYR_OPT
         dpk_kck_ok = self.pudpk.strength == self.SI_DPKCKR_DEFKICK
@@ -898,6 +902,8 @@ class InjSysPUModeHandler(_Devices, _Callback):
     @property
     def is_onaxis(self):
         """If configuration is OnAxis."""
+        if not self.connected:
+            return False
         dpk_evt_ok = self.trigdpk.source == self.trignlk.source
         dpk_dly_ok = self.trigdpk.delay_raw == \
             self.trignlk.delay_raw + self.SI_DPKCKR_DLYR_ONAXINC
--- a/siriuspy/siriuspy/devices/injsys.py
+++ b/siriuspy/siriuspy/devices/injsys.py
@@ -817,11 +817,7 @@ class InjSysPUModeHandler(_Devices, _Callback):
         """Switch to Accumulation."""
         self._abort.clear()
         self._update_status('Switching to PU Accumulation config...')
-        if not self.pudpk.interlock_ok:
-            self._update_status('ERR: DpK interlocks not ok. Aborted.')
-            return False
-        if not self.punlk.interlock_ok:
-            self._update_status('ERR: NLK interlocks not ok. Aborted.')
+        if not self._check_pu_interlocks():
             return False
         if self._abort.is_set():
             self._clr_flag_abort()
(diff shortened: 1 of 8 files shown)


siriuspy/siriuspy/devices/injsys.py
```diff
@@ -771,3 +773,279 @@ def _command_target(self, cmdtype):
...
+            return False
+
+        # set pulsed magnet pwrstate and pulse
+        proced = (
+            (self.pudpk.cmd_turn_on, 'turn DpK on.'),
+            (self.punlk.cmd_turn_on, 'turn NLK on.'),
+            (self.pudpk.cmd_turn_on_pulse, 'turn DpK pulse on.'),
+            (self.punlk.cmd_turn_on_pulse, 'turn NLK pulse on.'),
+        )
+        for fun, msg in proced:
```
murilobalves, line 928: maybe create a private method for this procedure, since it is repeated 3 times along the class

siriuspy/siriuspy/devices/injsys.py
```diff
@@ -771,3 +773,279 @@ def _command_target(self, cmdtype):
...
+            return False
+        if not self.punlk.interlock_ok:
+            self._update_status('ERR: NLK interlocks not ok. Aborted.')
+            return False
+        if self._abort.is_set():
+            self._clr_flag_abort()
+            return False
+
+        # if previously in on-axis, do delta angle x
+        if self.is_onaxis:
```
murilobalves, line 831: The block after this if statement also seems to be repeated along the code, except for the value of delta applied in posang.delta_angx. Maybe creating a method to do this process that takes as input the angle delta to be applied in posang might be good

siriuspy/siriuspy/devices/injsys.py
```diff
@@ -771,3 +773,279 @@ def _command_target(self, cmdtype):
...
+            (self.pudpk.pulse == PowerSupplyPU.PULSTATE.On)
+        nlk_off = (self.punlk.pwrstate == PowerSupplyPU.PWRSTATE.Off) or \
+            (self.punlk.pulse == PowerSupplyPU.PULSTATE.Off)
+        return dpk_evt_ok & dpk_dly_ok & dpk_kck_ok & dpk_on & nlk_off
+
+    def cmd_switch_to_onaxis(self):
+        """Switch to OnAxis."""
+        self._abort.clear()
+        self._update_status('Switching to PU OnAxis config...')
+        if not self.pudpk.interlock_ok:
```
murilobalves, line 957: these checks for interlocks seem to be the same in every `cmd_switch_to...` method. Maybe create a method like `_check_pu_interlocks` that does this checking.

murilobalves (changes requested):

siriuspy/siriuspy/devices/injsys.py
anacso17: Thanks @murilobalves , I pushed a commit trying to avoid repeated blocks of code. Done!

siriuspy/siriuspy/devices/injsys.py
anacso17: Done!

siriuspy/siriuspy/devices/injsys.py
anacso17: Done!

murilobalves (approved):

siriuspy/siriuspy/devices/injsys.py
```diff
@@ -771,3 +773,269 @@ def _command_target(self, cmdtype):
...
+        self._update_status('Switching to PU OnAxis config...')
+        if not self._check_pu_interlocks():
+            return False
+        if self._abort.is_set():
+            self._clr_flag_abort()
+            return False
+
+        # if not previously in on-axis, do delta angle x
+        if not self.is_onaxis:
+            delta = self.posang.delta_angx + self.TS_POSANG_DEFDELTA
```
fernandohds564, line 929: I'm not sure whether this is the right way to proceed, because inside `_do_delta_posang` you may reset the reference. In this case I think it is not appropriate to apply the old `delta_angx`, because it may have been incorporated to new reference.

siriuspy/siriuspy/devices/injsys.py
```diff
@@ -771,3 +773,269 @@ def _command_target(self, cmdtype):
...
+        if not self.punlk.interlock_ok:
+            self._update_status('ERR: NLK interlocks not ok. Aborted.')
+            return False
+        return True
+
+    def _do_delta_posang(self, delta):
+        if self.posang.need_ref_update:
+            self.posang.cmd_update_reference()
+        _time.sleep(InjSysPUModeHandler._DEF_SLEEP)
+        self.posang.delta_angx = delta
```
fernandohds564, line 972: maybe you could only do:

```python3
self.posang.delta_angx += delta
```

fernandohds564 (changes requested):

siriuspy/siriuspy/devices/injsys.py
anacso17: Thanks for noticing the bug! Fixed!

fernandohds564 (approved):

PR #827, opened 2022-07-27
Improve Devices

fernandohds564:
This PR contains:

- Add class to control all HLTiming triggers and events with methods to change triggers source and event delays without changing the associated triggers' total delays.
- Add EVG property called `machine_state`
- Add method in `FamBPMs` class to set all BPMs switching mode.
- Add protection in EVG setters of `continuous_state` and `injection_state` to only change values when needed.

### DIFF
--- a/siriuspy/siriuspy/devices/timing.py
+++ b/siriuspy/siriuspy/devices/timing.py
@@ -80,7 +80,7 @@ class EVG(_Device):
         """."""
         new_val = bool(value)
         if self.continuous_state != new_val:
-            self['ContinuousEvt-Sel'] = bool(value)
+            self['ContinuousEvt-Sel'] = int(new_val)
 
     @property
     def state_machine(self):
@@ -102,7 +102,7 @@ class EVG(_Device):
         """."""
         new_val = bool(value)
         if self.injection_state != new_val:
-            self['InjectionEvt-Sel'] = bool(value)
+            self['InjectionEvt-Sel'] = int(new_val)
 
     @property
     def injection_count_total(self):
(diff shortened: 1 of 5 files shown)


siriuspy/siriuspy/devices/timing.py
```diff
@@ -65,22 +73,36 @@ def bucketlist_sync(self):
...
-        return self['InjectionEvt-Sts']
+        return bool(self['InjectionEvt-Sts'])
 
     @injection_state.setter
     def injection_state(self, value):
         """."""
-        self['InjectionEvt-Sel'] = bool(value)
+        new_val = bool(value)
+        if self.injection_state != new_val:
+            self['InjectionEvt-Sel'] = bool(value)
```
anacso17, line 105: I think you can use `self['InjectionEvt-Sel'] = new_val`

siriuspy/siriuspy/devices/timing.py
```diff
@@ -65,22 +73,36 @@ def bucketlist_sync(self):
...
-        return self['ContinuousEvt-Sts']
+        return bool(self['ContinuousEvt-Sts'])
 
     @continuous_state.setter
     def continuous_state(self, value):
         """."""
-        self['ContinuousEvt-Sel'] = bool(value)
+        new_val = bool(value)
+        if self.continuous_state != new_val:
+            self['ContinuousEvt-Sel'] = bool(value)
```
anacso17, line 83: idem here

anacso17 (approved):

siriuspy/siriuspy/devices/timing.py
fernandohds564: actually it is better to write integer values to this PV.

anacso17 (approved):

PR #843, opened 2022-09-06
Add ID correctors family to lattice_survey

xresende:
We may have already discussed this but... shouldn't this lattice-survey script be in pymodels or in apsuite? after all, it does depend on pyaccel, pymodels... @anacso17 , @fernandohds564 , @murilobalves ? 
(I think the model table does belong to control-system, but a script to auto-generate it, or any other method, not necessarily.)


fernandohds564:
@xresende, I remember having this conversation and I also remember voting and arguing in favor of keeping this code in siriuspy. However, I don't remember the motivations for it. As for nowadays, I don't have a strong opinion about it, so for me we can change them if you guys prefer.

siriuspy/siriuspy/optics/lattice_survey.py
```diff
@@ -60,6 +60,11 @@ def _append_mag_data(filename, model, acc, label, section):
     mag_tps.extend(model.families.families_vertical_correctors())
     mag_tps.extend(model.families.families_sextupoles())
     mag_tps.extend(model.families.families_skew_correctors())
+    try:
```
fernandohds564, line 63: I prefer this style of coding:
```python3
if acc.lower().startswith('si'):
    mag_tps.extend(model.families.families_id_correctors()
```
It is shorter and self describable.

murilobalves:
I don't remember using the lattice-survey module directly. It seems that the other possible location where this code could be is pymodels. But I think you guys who have been using the module more often are more appropriate to decide this.

murilobalves (approved):

siriuspy/siriuspy/optics/lattice_survey.py
xresende: ok, done.

PR #896, opened 2023-01-27
Add Topup BiasFB

fernandohds564:
This PR main feature is the addition of a Bias Feedback to the InjCtrl IOC for topup operation.
The logic of the feedback was inspired in the class added in lnls-fac/apsuite#201. 

Besides this addition, a few other classes were changed:
 - Callback: I added a property called `has_callbacks`.
 - CurrInfoSI: the InjCurr PV will always be updated when Current3GeV from the booster is updated.
 - CurrInfoSI, Lifetime and InjCtrl: the increment of the `-Cmd` was delegated to the Driver class of the IOC. See lnls-sirius/machine-applications#266

### DIFF
--- a/siriuspy/siriuspy/currinfo/main.py
+++ b/siriuspy/siriuspy/currinfo/main.py
@@ -611,17 +611,12 @@ class SICurrInfoApp(_CurrInfoApp):
         timestamp_dq = _np.asarray(timestamp_dq)
         value_dq = _np.asarray(value_dq)
 
-        # check buffer not empty
-        if not timestamp_dq.size:
-            return
-
-        # check if there is valid current in Booster
-        if bo_curr < self.CURR_THRESHOLD:
-            return
-
-        # calculate efficiency
-        self._injcurr = value_dq.max() - value_dq.min()  # mA
-        self._injeff = 100*(self._injcurr/bo_curr) * self.HARMNUM_RATIO
+        self._injcurr = 0.0
+        self._injeff = 0.0
+        # check if buffer not empty and if there is valid current in Booster
+        if timestamp_dq.size and bo_curr >= self.CURR_THRESHOLD:
+            self._injcurr = value_dq.max() - value_dq.min()  # mA
+            self._injeff = 100*(self._injcurr/bo_curr) * self.HARMNUM_RATIO
 
         # calculate injected charge: 1e6 * mA / Hz = nC
         self._injcharge = 1e6*self._injcurr*self.HARMNUM/self._rffreq_pv.value
--- a/siriuspy/siriuspy/currinfo/main.py
+++ b/siriuspy/siriuspy/currinfo/main.py
@@ -615,7 +615,7 @@ class SICurrInfoApp(_CurrInfoApp):
         self._injeff = 0.0
         # check if buffer not empty and if there is valid current in Booster
         if timestamp_dq.size and bo_curr >= self.CURR_THRESHOLD:
-            self._injcurr = value_dq.max() - value_dq.min()  # mA
+            self._injcurr = value_dq.ptp()  # mA
             self._injeff = 100*(self._injcurr/bo_curr) * self.HARMNUM_RATIO
 
         # calculate injected charge: 1e6 * mA / Hz = nC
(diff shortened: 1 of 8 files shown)


siriuspy/siriuspy/injctrl/csdev.py
```diff
@@ -289,3 +297,192 @@ def get_injctrl_propty_database():
...
+        'MaxVoltage-SP': {
+            'type': 'float', 'value': -52, 'unit': 'V',
+            'prec': 1, 'lolim': -120, 'low': -120, 'lolo': -120,
+            'hilim': -30.0, 'high': -30.0, 'hihi': -30.0},
+        'MaxVoltage-RB': {
+            'type': 'float', 'value': -52, 'unit': 'V',
+            'prec': 1, 'lolim': -120, 'low': -120, 'lolo': -120,
+            'hilim': -30.0, 'high': -30.0, 'hihi': -30.0},
+
+        'ModelType-Sel': {
```
xresende, line 326: we should add units for similar PVs (`ModelMaxNumPts-SP/RB`, for example) in order for archiver viewer to be used more easily.

xresende (changes requested):

siriuspy/siriuspy/injctrl/csdev.py
fernandohds564: I don't understand @xresende. I have added units for this PV you commented. I just didn't add for `-Sel/-Sts` PVs. Do you think I should add for them too?

siriuspy/siriuspy/currinfo/main.py
```diff
@@ -611,17 +611,12 @@ def _update_injeff(self):
...
-            return
-
-        # calculate efficiency
-        self._injcurr = value_dq.max() - value_dq.min()  # mA
-        self._injeff = 100*(self._injcurr/bo_curr) * self.HARMNUM_RATIO
+        self._injcurr = 0.0
+        self._injeff = 0.0
+        # check if buffer not empty and if there is valid current in Booster
+        if timestamp_dq.size and bo_curr >= self.CURR_THRESHOLD:
+            self._injcurr = value_dq.max() - value_dq.min()  # mA
```
murilobalves, line 618: you can do the same with `value_dq.ptp()` to get peak-to-peak value, if you want

murilobalves (approved):

siriuspy/siriuspy/currinfo/main.py
fernandohds564: Nice! Done, thanks!

siriuspy/siriuspy/injctrl/csdev.py
xresende: I dont understand either. I meant `ModelMaxNumPts-SP/RB`, as I wrote. But right now I can see on this github.com PR interface that the units are there! I remember looking at the pair `ModelMaxNumPts-SP/RB` and not finding unit in neither of them. And you did not commit any changes... Well, most likely I overlooked something... sorry.

As to `Sel/Sts` PVs, I think we should also add units. Specially if the property is such that other devices might have identical or similar properties. Besides, adding units  you avoid archiver viewer having one axis for `Sel`and another one for `Sts`. I would suggest we copy the name of the property (`ModelType`, for example) to the PV unit field.

xresende (changes requested):

siriuspy/siriuspy/injctrl/csdev.py
fernandohds564: ok! I can do that!

fernandohds564: @xresende, I've done that. I'll explain how I did it in the main thread so it is more visible to others.

fernandohds564:
@xresende, as you requested, I added units to the `-Sel/Sts` and also to the `-Mon` PVs which are of type `enum`. I also added units to `int` PVs which are arrays of booleans to indicate interlock or ok/not ok states.

However I didn't follow your suggestion of copying the name of the PV to the unit. I thought it would be more descriptive to summarize all the possible states meaning in order, so the user know how to translate the numbers. For example, the `Off/On` PVs have unit `Off_On`.

The `int` PVs which are arrays of booleans follow the same idea, but the unit has a `2^_` in front of them to indicate that the value of the PV is a sum of powers of two.

@anacso17 , I tried to propagate this idea to the other PVs of the InjCtrl IOC. Please let me know what is your opinion about it.

xresende (approved):

RafaelLyra8 (approved):

anacso17 (approved):

PR #908, opened 2023-02-09
Fix AS-CurrInfo

fernandohds564:
This PR attempts to fix the connection problem with the Oscilloscope that freezes communications and the values of the PVs.

I'm not sure whether closing and reopening the connection will work though. We need to test this for an extended period to validate this solution.

maybe the right way to go would be to abandon `pyvisa` and use `socket` connections, just like the `oscilloscope` subpackage does.

What do you guys think about this @anacso17  and @xresende ?


xresende:
@fernandohds564 , I think it is worth trying reopening pyvisa connection, as you did here.
I was wondering whether there is a python library ready that abstracts the communication with this scope. What brand is it? tektronics/keysight? I suggested the same for andrei regarding keysight scopes but i do not know if he had the time to search it.

fernandohds564:
It is a keysight scope just like the one being used as filling pattern monitor, which means that the solutions implemented in the `Oscilloscope` subpackage would work here too.

anacso17 (approved):

xresende (approved):

murilobalves (approved):

PR #918, opened 2023-03-06
Add option of connecting with PostMorten PVs in BPM and FamBPMs.

fernandohds564:


murilobalves (approved):

siriuspy/siriuspy/devices/bpm.py
```diff
@@ -803,6 +845,15 @@ def cmd_sync_monit(self):
         self.monit_sync_enbl = 0
         return self._wait('Monit1TagEn-Sts', 0)
 
+    def _get_postmorten_prop(self, prop):
```
anacso17, line 848: wouldn't be better to use `_get_propname` or something that do not use "postmorten" in name?

xresende (approved):

siriuspy/siriuspy/devices/bpm.py
fernandohds564: right, done!

anacso17 (approved):

PR #922, opened 2023-03-09
Fix BPM and FamBPMs

fernandohds564:
- fix some bugs introduced in commit 1449303.
 - Start using recently created PV `ACQCount-Mon`, which is incremented every time an acquisition is finished, to indicate whether triggered acquisitions are ready to be assembled in an orbit. Thanks to @ericonr for implementing this PV!.


anacso17:
This PR depends on BPM IOC updates that will occur in Marche 13.

fernandohds564:
Yes, it does @anacso17. But your next deploy will happen only after that, right?

murilobalves (approved):

anacso17:
> Yes, it does @anacso17. But our next deploy will happen only after that, right?

Yes @fernandohds564 ! Just registering to let people that are not in other offline chats know :)

anacso17 (approved):

xresende (approved):

PR #923, opened 2023-03-09
Improve Normalizer

fernandohds564:
- Fix bug in initializer of TrimNormalizer.
 - Add optional attribute called `default_strengths_dipole` to normalizer classes to avoid having to define `strengths_dipole` on every conversion call.
 - Remove unused hiden module constants.

### DIFF
--- a/siriuspy/siriuspy/magnet/normalizer.py
+++ b/siriuspy/siriuspy/magnet/normalizer.py
@@ -20,8 +20,6 @@ if not _BETA_APPROXIMATION:
     _GAMMA_2_GEV = _mp.constants.electron_rest_energy * _mp.units.joule_2_GeV
 
 _MAGFUNCS = _mutil.get_magfunc_2_multipole_dict()
-_IS_DIPOLE = _re.compile(".*:[A-Z]{2}-B.*:.+$")
-_IS_FAM = _re.compile(".*[A-Z]{2}-Fam:[A-Z]{2}-.+$")
 _KCOEFF = _mp.constants.elementary_charge / \
           _mp.constants.light_speed / \
           _mp.constants.electron_mass / \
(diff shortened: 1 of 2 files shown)


anacso17 (approved):

siriuspy/siriuspy/magnet/normalizer.py
```diff
@@ -31,8 +29,10 @@
 class _MagnetNormalizer:
     """Base class for converting magnet properties: current and strength."""
 
-    def __init__(self, maname, magnet_conv_sign=-1):
+    def __init__(
+            self, maname, magnet_conv_sign=-1, default_strengths_dipole=None):
         """Class constructor."""
+        self.default_strengths_dipole = default_strengths_dipole  # [GeV]
```
xresende, line 35: @fernandohds564 , here comes again the question on class API. before we change to a new python version and resume discussions on our standard coding I suggest we try to keep consistency at least within each class separately. In this case the class uses hidden attributes and associated property and setter decorated methods.

siriuspy/siriuspy/magnet/normalizer.py
```diff
@@ -355,23 +370,24 @@ class TrimNormalizer(_MagnetNormalizer):
 
     TYPE = 'TrimNormalizer'
 
-    def __init__(self, maname, magnet_conv_sign=-1.0,
-                 **kwargs):
+    def __init__(self, maname, **kwargs):
```
xresende, line 373: was there a wrong conv sign ?! is this the init bug you mentioned?

xresende (changes requested):

murilobalves (approved):

siriuspy/siriuspy/magnet/normalizer.py
fernandohds564: Ok. I don't like the idea, but it's done!

xresende (approved):

PR #931, opened 2023-03-16
Timesys add DurationRaw PVs

fernandohds564:
The ideia of adding these PVs is to use them in the global config, so that we avoid the harmless but annoying rounding errors that come up when we try to set Duration PVs from an old configuration on the machine.

### DIFF
--- a/siriuspy/siriuspy/timesys/csdev.py
+++ b/siriuspy/siriuspy/timesys/csdev.py
@@ -511,13 +511,24 @@ def get_hl_trigger_database(hl_trigger, prefix=''):
     dbase['Src-Sts'] = _dcopy(dic_)
     dbase['Src-Sel'] = dic_
 
+    max_dur = 17e6
+    max_dur_raw = 2123400000
     dic_ = {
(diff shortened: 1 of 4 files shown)


murilobalves (approved):

anacso17 (approved):

siriuspy/siriuspy/timesys/csdev.py
```diff
@@ -511,23 +511,30 @@ def get_hl_trigger_database(hl_trigger, prefix=''):
     dbase['Src-Sts'] = _dcopy(dic_)
     dbase['Src-Sel'] = dic_
 
+    max_dur = 17e6
+    max_wid_raw = 2123400000
```
mmdonatti, line 515: `max_wid_raw` can be 2147483648 = 2^31 - 1

fernandohds564: thanks. done!

mmdonatti (approved):

xresende (approved):

PR #947, opened 2023-04-06
Adapt classes and devices to new FOFB controllers IOC version

anacso17:
This PR updates FOFB related classes according to new FOFB controllers IOC version:
- now we have RefOrbX and RefOrbY PVs, instead of only RefOrb
- equivalently, we have InvRespMatRowX and InvRespMatRowY, instead of InvRespMatRow
- CCEnable-RB PV was renamed to CCEnable-Sts

Also, this PR:
- separates FOFB IOC periodic checks in two threads, one for correctors and the other to controllers
- add FOFB controllers physical triggers PVs to as_diagnostics config

### DIFF
--- a/siriuspy/siriuspy/clientconfigdb/types/as_diagnostics.py
+++ b/siriuspy/siriuspy/clientconfigdb/types/as_diagnostics.py
@@ -1413,6 +1413,68 @@ _tune_pvs = [
     ['SI-Glob:DI-TuneProc-V:FiltType-Sel', 0, 0],
     ]
 
+_fofbctrls = [
+    'IA-01RaBPM:BS-FOFBCtrl',
+    'IA-02RaBPM:BS-FOFBCtrl',
+    'IA-03RaBPM:BS-FOFBCtrl',
+    'IA-04RaBPM:BS-FOFBCtrl',
+    'IA-05RaBPM:BS-FOFBCtrl',
+    'IA-06RaBPM:BS-FOFBCtrl',
+    'IA-07RaBPM:BS-FOFBCtrl',
+    'IA-08RaBPM:BS-FOFBCtrl',
+    'IA-09RaBPM:BS-FOFBCtrl',
+    'IA-10RaBPM:BS-FOFBCtrl',
+    'IA-11RaBPM:BS-FOFBCtrl',
+    'IA-12RaBPM:BS-FOFBCtrl',
+    'IA-13RaBPM:BS-FOFBCtrl',
+    'IA-14RaBPM:BS-FOFBCtrl',
+    'IA-15RaBPM:BS-FOFBCtrl',
+    'IA-16RaBPM:BS-FOFBCtrl',
+    'IA-17RaBPM:BS-FOFBCtrl',
+    'IA-18RaBPM:BS-FOFBCtrl',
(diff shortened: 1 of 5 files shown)


xresende (approved):

fernandohds564 (approved):

RafaelLyra8 (approved):

murilobalves (approved):

siriuspy/siriuspy/clientconfigdb/types/as_diagnostics.py
```diff
@@ -1413,6 +1413,68 @@ def get_dict():
...
+    'IA-13RaBPM:BS-FOFBCtrl',
+    'IA-14RaBPM:BS-FOFBCtrl',
+    'IA-15RaBPM:BS-FOFBCtrl',
+    'IA-16RaBPM:BS-FOFBCtrl',
+    'IA-17RaBPM:BS-FOFBCtrl',
+    'IA-18RaBPM:BS-FOFBCtrl',
+    'IA-19RaBPM:BS-FOFBCtrl',
+    'IA-20RaBPM:BS-FOFBCtrl',
+]
+_fofb_propts = [
```
fernandohds564, line 1438: @anacso17, are these triggers used only for diagnostics purposes? What are they used for? 

PS.: Maybe @ericonr could also answer these questions.

siriuspy/siriuspy/clientconfigdb/types/as_diagnostics.py
```diff
@@ -1413,6 +1413,68 @@ def get_dict():
...
+    [':TRIGGER1RcvLen-SP', 1, 0.0],
+    [':TRIGGER1TrnLen-SP', 1, 0.0],
+    [':TRIGGER2Dir-Sel', 0, 0.0],
+    [':TRIGGER2DirPol-Sel', 1, 0.0],
+    [':TRIGGER2RcvLen-SP', 1, 0.0],
+    [':TRIGGER2TrnLen-SP', 1, 0.0],
+    [':TRIGGER3Dir-Sel', 0, 0.0],
+    [':TRIGGER3DirPol-Sel', 1, 0.0],
+    [':TRIGGER3RcvLen-SP', 1, 0.0],
+    [':TRIGGER3TrnLen-SP', 1, 0.0],
```
ericonr, line 1454: Trigger 3 is the output for orbit interlock, which isn't used by the FOFB controller. I believe we can set Dir-Sel to 1.

siriuspy/siriuspy/clientconfigdb/types/as_diagnostics.py
ericonr: We need to configure them so they don't affect the physical trigger buses, which are used mainly by the BPMs, at least for now.

We plan on using one of the lines for system identification; I don't know in which configuration that would fit, though.

siriuspy/siriuspy/clientconfigdb/types/as_diagnostics.py
anacso17: Hi @ericonr , the values in this file are just default/example values. The values that will be stored in the configuration will be read from the PV values. So the procedure is to configure all triggers accordingly and then save a configuration :)

PR #950, opened 2023-04-17
Update idff subpackage and add first version of IDFF IOC

xresende:
- Update si_idff
- Update IDFF devices according to new si_idff configuration
- Add a first version of IDFF IOC


fernandohds564 (approved):

siriuspy/siriuspy/clientconfigdb/types/si_idff.py
```diff
@@ -22,16 +22,24 @@ def get_dict():
...
 #   - 'polarizations': its value is a dict with
-#       ('horizontal', 'vertical', ...) light polarization configs of the ID
-#       as keys and dicts with kparameter and correctors
+#       ('none', 'horizontal', 'vertical', ...) light polarization configs of
+#       the ID as keys and dicts with kparameter and correctors
 #       IDFF tables as lists.
+#
+#   obs:
+#   1) pparameter: pvname that varies polarization configurations
+#   2) kparameter: pvname that varties k (photon energy) configurations
```
anacso17, line 35: typo: `varties`

siriuspy/siriuspy/devices/idff.py
```diff
@@ -101,37 +130,108 @@ def calculate_setpoints(self, polarization):
...
+
+            # check 'pparameter'
+            if 'pparameter' not in table:
+                raise ValueError(
+                    'Missing pparameter in polarization configuration.')
+
+            # check 'kparameter'
+            if 'kparameter' not in table:
+                raise ValueError(
+                    'Missing pparameter in polarization configuration.')
```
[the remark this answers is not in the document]
xresende: done

Gabrielrezende-asc (approved):

siriuspy/siriuspy/idff/main.py
```diff
@@ -0,0 +1,226 @@
...
+            'LoopFreq-RB': self._loop_freq,
+            'ControlQS-Sel': self._control_qs,
+            'ControlQS-Sts': self._control_qs,
+            'ConfigName-SP': self._config_name,
+            'ConfigName-RB': self._config_name,
+            'Polarization-Mon': self._polarization,
+        }
+        for pvn, val in pvn2vals.items():
+            self.run_callbacks(pvn, val)
+        self._load_config(self._config_name)
```
xresende, line 65: @anacso17 , why did you include here the loading of the IDFF? as it is now it will always fail for config name 'none' in the IOC initialization, right? do you intend to change the default config name  from 'none' to another configuration in the future? if it is the case I suggest either moving this loading to a separate initialization section of renaming the method from `init_database` to something more general, `init`.

siriuspy/siriuspy/idff/main.py
```diff
@@ -0,0 +1,226 @@
...
+            status = self.map_pv2write[reason](value)
+            _log.info('%s Write for: %s --> %s',
+                      str(status).upper(), reason, str(value))
+            return status
+        _log.warning('PV %s does not have a set function.', reason)
+        return False
+
+    def set_loop_state(self, value):
+        """Set loop state."""
+        if not 0 <= value < len(_ETypes.OPEN_CLOSED):
```
xresende, line 100: does pcaspy not take care already of this checking? I dont recall.
if not, I think we should also update `Log-Mon` indicating the error.

siriuspy/siriuspy/idff/main.py
```diff
@@ -0,0 +1,226 @@
...
+        self._loop_state = value
+        act = ('En' if value else 'Dis')
+        self._update_log(f'Feedforward loop {act}abled.')
+        self.run_callbacks('LoopState-Sts', value)
+        return True
+
+    def set_loop_freq(self, value):
+        """Set loop frequency."""
+        if not 1e-3 <= value < 60:
+            return False
```
xresende, line 111: I think we should also update `Log-Mon` indicating the error. (or is it already a standard in other IOCs just to ignore the situation?)

siriuspy/siriuspy/idff/main.py
```diff
@@ -0,0 +1,226 @@
...
+        self._loop_freq = value
+        self._update_log(f'Loop frequency updated to {value:.2f}Hz.')
+        self.run_callbacks('LoopFreq-RB', value)
+        return True
+
+    def set_control_qs(self, value):
+        """Set whether to include QS or not in feedforward."""
+        if not 0 <= value < len(_ETypes.DSBL_ENBL):
+            return False
```
xresende, line 120: same as before

siriuspy/siriuspy/idff/main.py
```diff
@@ -0,0 +1,226 @@
...
+        return True
+
+    def set_config_name(self, value):
+        """Set configuration name."""
+        if self._loop_state == self._const.LoopState.Closed:
+            self._update_log('ERR:Open loop before changing configuration.')
+            return False
+
+        if not self._load_config(value):
+            return False
```
xresende, line 134: I think we should definitely update `Log-Mon` indicating the error.

xresende: oh ok, I see... `Log-Mon` is updated within _load_config...

siriuspy/siriuspy/idff/main.py
```diff
@@ -0,0 +1,226 @@
...
+            _log.error(msg[6:])
+        elif 'WARN' in msg:
+            _log.warning(msg[5:])
+        else:
+            _log.info(msg)
+        self.run_callbacks('Log-Mon', msg)
+
+    # ----- auto save methods -----
+
+    def read_autosave_file(self):
```
xresende, line 164: I never liked the idea of autosave implemented in local filesystem as the initial state of the IOC depends on where it runs (production in container/storage vs local test instantiations). I always thought configdb should we used for this (as this service should always be available!).

But if we will keep with local filesystem read/write, should not these functions be genralized and moved to a general section in siriuspy, so that all IOCs use the same read/save parameterized methods?

siriuspy/siriuspy/idff/main.py
xresende: I see, @anacso17 , you are updating _config_name within `read_autosave_file` invoked previously. Then I suggest renaming `init_database` or loaginf the IDFF config in a separate section/method.

siriuspy/siriuspy/idff/main.py
anacso17: for enum PVs pcaspy do not take care of the checking.
I don't think it is necessary for this case to log the error, but I will do.

siriuspy/siriuspy/idff/main.py
anacso17: In other IOCs I usually do not log these type of setpoints, I usually log as an error setpoints that are valid values but could not be implemented for any operation point of the IOC

siriuspy/siriuspy/idff/main.py
```diff
@@ -0,0 +1,226 @@
...
+        self._loop_state = _Const.DsblEnbl.Dsbl
+        self._loop_freq = _Const.DEFAULT_LOOP_FREQ
+        self._control_qs = _Const.DsblEnbl.Dsbl
+        self._polarization = 'none'
+        self._config_name = ''
+        self.read_autosave_file()
+
+        # connections
+        # use pyepics recommendations for threading
+        _epics.ca.use_initial_context()
```
xresende, line 33: do we need to invoke `use_initial_context` here? it is already being invoked within `CAThread.run`that is called in `CAThread.start`:

```Python3

class CAThread(threading.Thread):
    """
    Sub-class of threading.Thread to ensure that the
    initial CA context is used.
    """
    def run(self):
        use_initial_context()
        threading.Thread.run(self)

```

siriuspy/siriuspy/idff/main.py
anacso17: Just like other (not pcaspy) IOCs, autosave is a feature that is usuful for a set of devices, so the user do not need to resend all configurations when IOC is restarted. 

> should not these functions be genralized and moved to a general section in siriuspy, so that all IOCs use the same read/save parameterized methods?

maybe, I see we have been using autosave for some IOCs, but I don't think this is the scope of this PR. Maybe we could unify the autosave feature in a future PR.

siriuspy/siriuspy/idff/main.py
anacso17: I usually call use_initial_context before creating PVs in the App class, but maybe we don't need it

siriuspy/siriuspy/clientconfigdb/types/si_idff.py
xresende: done

siriuspy/siriuspy/devices/idff.py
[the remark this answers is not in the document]
xresende: done

siriuspy/siriuspy/idff/main.py
xresende: yes, I know it is a feature. but in my view it is a design flaw. there should be an epics service for this. in our case we have configdb

siriuspy/siriuspy/idff/main.py
anacso17: done!

siriuspy/siriuspy/idff/main.py
anacso17: done!

murilobalves (approved):

PR #955, opened 2023-04-24
Use hostnames instead of IPs for oscilloscopes

anacso17:
This PR updates oscilloscope subspackage to use scopes hostnames instead of IPs.


murilobalves (approved):

xresende (approved):

siriuspy/siriuspy/oscilloscope/keysight.py
```diff
@@ -183,8 +174,8 @@ def stats_get_data(self):
             data = self.stats_acquire()
             print('Total acquisition time:', _time.time() - tini)
         except Exception:
-            print("Unexpected error:", _sys.exc_info()[0])
-            print('Close connetion by exception')
+            print('Close connection by exception')
+            raise
```
fernandohds564, line 178: esse `raise` sozinho levanta o que?

siriuspy/siriuspy/oscilloscope/keysight.py
```diff
@@ -142,8 +144,9 @@ def wfm_get_data(self, channel=None):
             print('Total acquisition time:', _time.time() - tini)
             # self.send_command(b":WAVeform:STReaming ON\n", get_res=False)
         except Exception:
-            print("Unexpected error:", _sys.exc_info()[0])
-            print('Close connetion by exception')
+            print('Close connection by exception')
+            raise
```
fernandohds564, line 148: same here.

fernandohds564: Actually, raising the error again changes the interface of the method. This could crash the LI/TB/TS CurrInfo IOCs that use this class. Why did you decide to raise the error again?

fernandohds564 (changes requested):

siriuspy/siriuspy/oscilloscope/keysight.py
anacso17: It raises the same Exception, without changes

fernandohds564: nice, I didn't know that!

But I think the problem with is in the comment below.

siriuspy/siriuspy/oscilloscope/keysight.py
anacso17: Maybe in these IOCs we should catch the exception?

anacso17: Actually, it seems the IOC does not use this function

fernandohds564: Nice, ok then!

fernandohds564 (approved):

PR #968, opened 2023-05-25
Make SOFB aware of AMC lock state

fernandohds564:


siriuspy/siriuspy/sofb/main.py
```diff
@@ -942,6 +950,10 @@ def _do_auto_corr(self):
                 self._update_log(msg)
                 _log.info(msg)
 
+        if self._loop_state == self._csorb.LoopState.Closed:
+            self._loop_state = self._csorb.LoopState.Open
+            self.run_callbacks('LoopState-Sel', 0)
```
anacso17, line 955: maybe we could use `self.run_callbacks('LoopState-Sel', self._csorb.LoopState.Open)`, just to avoid hardcoded constants

anacso17 (approved):

xresende (approved):

murilobalves (approved):

siriuspy/siriuspy/sofb/main.py
```diff
@@ -840,6 +851,11 @@ def _do_auto_corr(self):
                 self._update_log(msg)
                 _log.info(msg)
                 break
+            if not self.is_amc_locked:
+                msg = 'ERR: Cannot Correct, at least one AMC is not locked!'
+                self._update_log(msg)
+                _log.info(msg)
+                break
```
ericonr, line 858: When we had connection issues, this error message was somewhat misleading, because it suggested that the timing wasn't locked, when in fact the issue was that the PV wasn't connected.

Do you think it would be easy to differentiate between the two issues in the log message?

If we lost connection to one of the crates, it still makes sense to open the SOFB loop, so the overall behavior should remain, I think

fernandohds564: Right, done!

ericonr (approved):

PR #970, opened 2023-05-25
Add device BLPPSCtrl

xresende:
Add BLPPSCtrl device (beamline personal protection system control) to open beamlines (CAX)

### DIFF
--- a/siriuspy/siriuspy/devices/blctrl.py
+++ b/siriuspy/siriuspy/devices/blctrl.py
@@ -1,9 +1,9 @@
 """Beamline Control."""
 
-from .device import Device as _Device
+from .device import DeviceNC as _DeviceNC
 
 
-class BLCtrl(_Device):
+class BLCtrl(_DeviceNC):
     """Beamline Control."""
 
     class DEVICES:
--- a/siriuspy/siriuspy/devices/blctrl.py
+++ b/siriuspy/siriuspy/devices/blctrl.py
@@ -3,7 +3,7 @@
 from .device import DeviceNC as _DeviceNC
 
 
-class BLCtrl(_DeviceNC):
+class BLPPSCtrl(_DeviceNC):
     """Beamline Control."""
 
     class DEVICES:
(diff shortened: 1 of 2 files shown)


fernandohds564 (approved):

siriuspy/siriuspy/devices/blctrl.py
```diff
@@ -0,0 +1,85 @@
+"""Beamline Control."""
+
+from .device import Device as _Device
+
+
+class BLCtrl(_Device):
```
anacso17, line 6: maybe we could use a name that refers to PPS system, this name is a little bit generic

siriuspy/siriuspy/devices/blctrl.py
```diff
@@ -0,0 +1,85 @@
+"""Beamline Control."""
+
+from .device import Device as _Device
```
anacso17, line 3: shouldn't it be derivated from DeviceNC?

siriuspy/siriuspy/devices/blctrl.py
xresende: I am not sure what else in the CAX will be useful and implemented in this device, that is why I left generic. Another option is to break into many classes.

siriuspy/siriuspy/devices/blctrl.py
xresende: done

siriuspy/siriuspy/devices/blctrl.py
fernandohds564: I like this last approach. Or at least include the prefix `pps_` in the property names.

anacso17: I also prefer the last approach

xresende: class renamed to `BLPPSCtrl`

anacso17 (approved):

murilobalves (approved):

Gabrielrezende-asc (approved):

PR #978, opened 2023-06-15
Fix device papu50

xresende:


Gabrielrezende-asc (approved):

siriuspy/siriuspy/devices/ids.py
```diff
@@ -300,6 +291,20 @@ def cmd_move_disable(self, timeout=None):
         """Command to disable and break ID phase and gap movements."""
         return self.cmd_move_phase_disable(timeout=timeout)
 
+    # --- cmd_wait
+
+    def cmd_wait_while_busy(self, timeout=None):
+        """Command wait within timeout while ID control is busy."""
+        return True
```
murilobalves, line 298: is this method "to be implemented"?

siriuspy/siriuspy/search/id_search.py
```diff
@@ -110,8 +110,12 @@ def conv_idname_2_idff(idname):
     def conv_idname_2_pparameter_propty(idname):
         """."""
         idff = IDSearch.conv_idname_2_idff(idname)
-        pvname = _SiriusPVName(idff['pparameter'])
-        return pvname.propty if pvname is not None else None
+        pparameter = idff['pparameter']
```
murilobalves, line 113: It will always be the case that 'pparameter' is a key of the idff dictionary?

siriuspy/siriuspy/devices/ids.py
```diff
@@ -300,6 +291,20 @@ def cmd_move_disable(self, timeout=None):
...
+
+    def cmd_wait_for_movement(self, timeout=None):
+        """Command wait until movement starts or timeout."""
+        time_init = _time.time()
+        while not self.is_moving:
```
murilobalves, line 303: is this right? It seems that `while self.is_moving` is the correct condition. 

Also, is the prefix `cmd_` appropriated for what this method is doing? I would say it is just a wait method, not setting any command for the device. This also happens for other wait methods in this device. On other devices we used `wait_something` to name these wait methods.

murilobalves (changes requested):

siriuspy/siriuspy/devices/ids.py
xresende: this is right, @murilobalves . the idea is to have an utility method we can invoke right after `move_start` so that we can guarantee, within timeout, that the "moving" state has already moved.

as for `cmd` prefix, I wondered myself. it indeed does not send commands to the device but I thought since it is used in association with other cmd commands, I grouped them together by prefix. but you are right, it is somewhat artificial... another option is to include this waiting algorithm as an option within the `cmd_move_phase_start`.

method `cmd_wait_while_busy` (which is material only for the EPU case) has the same issue about the prefix `cmd_` since it does not send commands, but just read status after othe `cmd` methods...

siriuspy/siriuspy/devices/ids.py
xresende: no. this method is immaterial for PAPU but makes sense for derived class EPU. I included this method in PAPU so that I did not hake to overwrite other methods for EPU.

siriuspy/siriuspy/devices/ids.py
xresende: done

anacso17 (approved):

murilobalves (approved):

siriuspy/siriuspy/search/id_search.py
xresende: yes, @murilobalves . for those IDs with only one type of polarization, for which pparameter is irrelevant, the valur of the key is always `None`.

PR #979, opened 2023-06-19
Fix auto_monitor settings in Device class and bugs in FOFB

anacso17:
Change auto_mon argument to a more accurate name, auto_monitor_mon, and add new argument auto_monitor to make it possible to configure all PV types.


siriuspy/siriuspy/devices/device.py
```diff
@@ -151,9 +180,11 @@ def _create_pvs(self, devname):
         pvs = dict()
         for propty in self._properties:
             pvname = self._get_pvname(devname, propty)
-            # avoid keeping auto_monitor enabled for -Mon PVs as they usually
-            # have a high update rate which can generate a lot of CPU load
-            auto_monitor = self._auto_mon and not pvname.endswith('-Mon')
+            auto_monitor = self._auto_monitor
+            if pvname.endswith('-Mon'):
+                auto_monitor = self._auto_monitor_mon
+            elif pvname.endswith('Data'):
```
fernandohds564, line 186: PVs ending with `'Data'` are not part of the SIRIUS naming convention and I think we should avoid creating the property `auto_monitor_data` unless strictly necessary.

As far as I understood from the code, the `'Data'` PVs are of the type `'-Mon'`, aren't they?
In this case, I couldn't we avoid the creation of the attribute with the following code?:
```python3
            if pvname.endswith(('-Mon', 'Data')):
                auto_monitor = self._auto_monitor_mon
```

fernandohds564 (changes requested):

siriuspy/siriuspy/devices/device.py
anacso17: agreed! done!

fernandohds564 (approved):

murilobalves (approved):

PR #989, opened 2023-06-28
Remove PSSOFB from SOFB

fernandohds564:
Tested with Beam.

Tests showed we cannot operate at 10 Hz without PSSOFB due to a very large latency of the correctors when operated via IOC. We do not understand this latency yet.
However, we (@xresende and I) evaluated that operation at 10 Hz is not needed anymore and we could run the loop at 5 Hz or less.

We tested the loop operation with 5 Hz and it worked well.


siriuspy/siriuspy/sofb/correctors.py
```diff
@@ -553,61 +530,13 @@ def apply_kicks(self, values):
         _log.info(strn.format('send evt:', 1000*(time4-time3)))
 
         # Wait for references to be updated
-        self._timed_out(values, mode='applied')
+        if WAIT_CORRS and self._timed_out(values, mode='applied'):
+            pass
         time5 = _time.time()
         _log.info(strn.format('check applied:', 1000*(time5-time4)))
         _log.info('    TIMEIT: END')
         return 0
```
anacso17, line 495: maybe this way:
```
if WAIT_CORRS:
    self._timed_out(values, mode='applied')
    time5 = _time.time()
    _log.info(strn.format('check applied:', 1000*(time5-time4)))
_log.info('    TIMEIT: END')
return 0
```
?

fernandohds564: done!

murilobalves (approved):

siriuspy/siriuspy/sofb/csdev.py
```diff
@@ -791,18 +790,6 @@ def get_sofb_database(self, prefix=''):
             'LoopPerfTimGetOStd-Mon': {
                 'type': 'float', 'value': 0, 'unit': 'ms', 'prec': 1,
                 'lolim': -1, 'hilim': 100},
-            'LoopPerfTimGetKMax-Mon': {
```
xresende, line 794: @fernandohds564 , @anacso17 , we should remember to take all these removed PVs from archiver.

xresende (approved):

ericonr (approved):
I don't have a deep understanding of the code, but couldn't find any great issues while skimming the commits.

Has the power supply latency been isolated as a problem in the power supply IOCs themselves, or could it be due to the great number of connections established by the python code in SOFB? Let me know if any aid is necessary when investigating these issues (or in implementing something like sofb-mon, which we currently use for the orbit, for the power supplies).

fernandohds564:
@ericonr, it seems to be related to the PSs IOCs or below, because the SP and Ref PVs of the currents were  not matching during the test. This means that SOFB was being able to write and the IOCs received the values, but the power supply itself was lagging.

xresende (approved):

anacso17 (approved):

murilobalves (approved):

vellosok75 (approved):

PR #991, opened 2023-06-30
Add PsMtn BPM trigger to global_config.

fernandohds564:
This PR : 
- adds high level PsMtn BPM trigger
- fixes a bug in HLTiming IOC that was causing erroneous Evt number parsing
- removes from as_diagnostics configuration the AMCFPGAEVR PVs that are controlled by HL triggers

### DIFF
--- a/siriuspy/siriuspy/timesys/ll_classes.py
+++ b/siriuspy/siriuspy/timesys/ll_classes.py
@@ -678,7 +678,7 @@ class _BASETRIG(_BaseLL):
             dic_ = {'Src': n}
         else:
             n -= offset
-            evt = int(_TIConst.EvtHL2LLMap[pname][-2:])
+            evt = int(_TIConst.EvtHL2LLMap[pname][3:])
             dic_ = {'Src': n, 'Evt': evt}
         if 'SrcTrig' in self._dict_convert_prop2pv.keys():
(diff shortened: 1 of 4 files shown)


mmdonatti (approved):

anacso17 (approved):

xresende (approved):

mmdonatti (approved):

murilobalves (approved):

siriuspy/siriuspy/timesys/ll_classes.py
```diff
@@ -678,7 +678,7 @@ def _set_source(self, value):
             dic_ = {'Src': n}
         else:
             n -= offset
-            evt = int(_TIConst.EvtHL2LLMap[pname][-2:])
+            evt = int(_TIConst.EvtHL2LLMap[pname][3:])
```
ericonr, line 681: Could we use some strip function for letters and symbols and leave just numbers instead? That seems like it would document the intent better.

anacso17: done!

ericonr (approved):
Approving the PR, but I noticed an issue with the high level event name: it should be PsMtm, since "post mortem" is spelled with an "M" at the end, not an "N".

PR #998, opened 2023-07-20
Add scrapers device

murilobalves:
- The properties related to the ScraperH Outer and Inner slits were renamed to be Left and Right slits. In that way the concept of outer and inner will always be with respect to the outer and inner side of the vacuum chamber. 

- @anacso17 We were discussing that it would be nice if the ScraperH slit names were changed in the IOC to be `Outer -> Left` and `Inner -> Right`. The ScraperV slits are ok as `Top` and `Bottom`.

- The setter of slit positions was implemented as a separated method, for example `move_bottom_slit`, not as a setter of the property `bottom_slit_pos`. Setting a value to `SlitPos-SP` will make the scraper slit actually move to that position, which might be problematic if done recklessly. A separated method with `move` in the name should call the user attention that this method will trigger a move action to the device.


siriuspy/siriuspy/devices/scraper.py
```diff
@@ -0,0 +1,311 @@
...
+    def left_slit_control_prefix(self):
+        """."""
+        return self['OuterMotionCtrl-Cte']
+
+    @property
+    def right_slit_control_prefix(self):
+        """."""
+        return self['InnerMotionCtrl-Cte']
+
+    def cmd_move_left_slit(self, value):
```
fernandohds564, line 152: methods starting with `cmd_` are reserved to act in `-Cmd` PVs, hence they don't accept an input value.
Please, consider changing the prefix of these methods to something else. I think that only `move_left_slit` is ok.

siriuspy/siriuspy/devices/scraper.py
```diff
@@ -0,0 +1,311 @@
...
+
+    def __init__(self):
+        """."""
+        # call base class constructor
+        _all_properties = _ScraperDev._properties + ScraperH._properties
+        super().__init__(
+            devname=_ScraperDev.DEVICES.H, properties=_all_properties)
+
+    @property
+    def left_slit(self):
```
fernandohds564, line 85: I think it would be more intuitive if these properties ended with a `_pos` to indicate that it returns the position of the slits.

siriuspy/siriuspy/devices/scraper.py
```diff
@@ -0,0 +1,311 @@
...
+        [Outer, Inner] | (x=0) | [Inner, Outer]
+        """
+        return [self['InnerSlitInnerLim-RB'], self['InnerSlitOuterLim-RB']]
+
+    @right_slit_limits.setter
+    def right_slit_limits(self, limits):
+        """Set right slit limits.
+
+        Left Slit (x>0)             Right Slit (x<0)
+        [Outer, Inner] | (x=0) | [Inner, Outer]
```
fernandohds564, line 137: :warning:  :banana: :face_with_head_bandage:

fernandohds564 (changes requested):
I think it is important to have methods to wait the slits finish moving. Maybe only one method for both slits is enough. I think the interface could be something like:
```python3
def wait_slits_finish_moving(self, timeout=10) -> bool:
```

anacso17:
> @anacso17 We were discussing that it would be nice if the ScraperH slit names were changed in the IOC to be Outer -> Left and Inner -> Right. The ScraperV slits are ok as Top and Bottom.

@aguinanc could we rename the ScraperH slit names PVs (again :sweat_smile:) ?

siriuspy/siriuspy/devices/scraper.py
```diff
@@ -0,0 +1,311 @@
...
+
+    @left_slit_limits.setter
+    def left_slit_limits(self, limits):
+        """Set left slit limits.
+
+        Left Slit (x>0)             Right Slit (x<0)
+        [Outer, Inner] | (x=0) | [Inner, Outer]
+        """
+        self['OuterSlitInnerLim-SP'] = limits[0]
+        self['OuterSlitOuterLim-SP'] = limits[1]
```
anacso17, line 121: there is an inconsistency here: 
getter returns outer and inner
setter receives inner and outer

anacso17: (assuming the bug pointed out above has been fixed) why are the limits sorted in descending order?

siriuspy/siriuspy/devices/scraper.py
murilobalves: done!

siriuspy/siriuspy/devices/scraper.py
murilobalves: done!

siriuspy/siriuspy/devices/scraper.py
murilobalves: changed!

siriuspy/siriuspy/devices/scraper.py
murilobalves: Yes that was a bug, thanks Ana. 

The definition was confusing indeed, as Fernando also pointed in the comment above. Now the limits are always `[inner, outer]`, where `inner` refers to the minimum distance from the slit to the beam and `outer` refers to the maximum distance. If we ignore the sign of values (inner and bottom slits the values are typically negative), now the limits are sorted in ascending order. Please take a look in the updated code.

siriuspy/siriuspy/devices/scraper.py
```diff
@@ -0,0 +1,283 @@
...
+        ret1 = self._wait(propty=pv_names[0], value=1, timeout=timeout)
+        if not ret1:
+            return ret1
+        timeout -= _time.time() - t0
+        ret2 = self._wait(propty=pv_names[1], value=1, timeout=timeout)
+        return ret2
+
+    @staticmethod
+    def _check_limits(limits):
+        if abs(limits[0]) > abs(limits[1]):
```
fernandohds564, line 72: this won't work every time. I think the following method will:
```python3
    @staticmethod
    def _check_limits(lims, is_negative=False):
        if (lims[1] == lims[0]) or ((lims[1] - lims[0] < 0) ^ is_negative):
            stg = 'larger' if is_negative else 'smaller'
            raise ValueError(f'Inner limit must be {stg:s} than outer limit.')
```

fernandohds564 (changes requested):

siriuspy/siriuspy/devices/scraper.py
murilobalves: Well noted, I implemented your suggestion, thanks!

fernandohds564 (approved):

Gabrielrezende-asc (approved):

anacso17 (approved):

xresende (approved):

PR #1008, opened 2023-08-03
TIM.BUG: Fix length of some char array PVs.

fernandohds564:
@anacso17 informed me that some char array PVs from the high level timing IOCs were causing instabilities on the PV Gateway. She told me the problem was related to PVs whose value length exceeded the `'count'`  of the PV. Namely, the following PVs:
```
SI-Glob:TI-Mags-Corrs:CtrldChannels-Cte
SI-Glob:TI-Mags-QTrims:CtrldChannels-Cte
```

I noticed the definition of the above mentioned field was hard-coded.
I changed it to be defined according to the real length of each value.

### DIFF
--- a/siriuspy/siriuspy/timesys/csdev.py
+++ b/siriuspy/siriuspy/timesys/csdev.py
@@ -605,15 +605,17 @@ def get_hl_trigger_database(hl_trigger, prefix=''):
     dbase['InInjTable-Mon'] = {
         'type': 'enum', 'enums': _et.ININJTAB, 'value': 0}
 
+    # NOTE: we need to add plus 1 to the PVs count due to some unexpected
+    # behavior of pcaspy
     labs = '\n'.join(Const.HLTrigStatusLabels)
     dbase['StatusLabels-Cte'] = {
-        'type': 'char', 'count': len(labs), 'value': labs}
+        'type': 'char', 'count': len(labs)+1, 'value': labs}
 
     ll_trigs = '\n'.join(ll_trig_names)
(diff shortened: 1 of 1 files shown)


anacso17 (approved):

murilobalves (approved):

siriuspy/siriuspy/timesys/csdev.py
```diff
@@ -605,15 +605,15 @@ def get_hl_trigger_database(hl_trigger, prefix=''):
     dbase['InInjTable-Mon'] = {
         'type': 'enum', 'enums': _et.ININJTAB, 'value': 0}
 
+    labs = '\n'.join(Const.HLTrigStatusLabels)
     dbase['StatusLabels-Cte'] = {
-        'type': 'char', 'count': 1000,
-        'value': '\n'.join(Const.HLTrigStatusLabels)
-        }
+        'type': 'char', 'count': len(labs) + 10, 'value': labs}
```
ericonr, line 610: Why are you adding 10 to them? Would it make sense to add a comment explaining it?

ericonr (approved):
This looks good!

If you'd like to be sure the element count is correct, you can check with a python client that `count` is correct:

```
>>> from epics import PV
>>> p = PV('SI-Glob:TI-Mags-QTrims:CtrldChannels-Cte')
>>> p 
<PV 'SI-Glob:TI-Mags-QTrims:CtrldChannels-Cte', count=5840/5000, type=time_char, access=read-only>
```

xresende (approved):

PR #1022, opened 2023-09-05
Add device BLM

xresende:


siriuspy/siriuspy/search/blm_search.py
```diff
@@ -0,0 +1,75 @@
...
+        """Return BLM name from device name."""
+        for blmname in BLMSearch._blmname2data:
+            sec, _, chn = BLMSearch._blmname2data[blmname]
+            devname_ = _SiriusPVName(sec + ':' + 'CO-Counter-' + chn)
+            if devname_ == devname:
+                return blmname
+        raise ValueError(f'Invalid devname {devname}')
+
+    @staticmethod
+    def conv_blmname_2_sensor_pvname(blmname):
```
anacso17, line 64: we don't usually have methods like this in search, do we need this one? I think pvnames could be listed only in devices

siriuspy/siriuspy/search/blm_search.py
```diff
@@ -0,0 +1,75 @@
...
+        """Return PV name of a BLM."""
+        devname = BLMSearch.conv_blmname_2_devname(blmname)
+        pvname = devname.substitute(propty='Count-Mon')
+        return pvname
+    
+    @staticmethod
+    def conv_blmname_2_timebase_pvname(blmname):
```
anacso17, line 71: idem here

siriuspy/siriuspy/search/blm_search.py
```diff
@@ -0,0 +1,75 @@
...
+    @staticmethod
+    def conv_blmname_2_position(blmname):
+        """Return the position of the BLM with sector."""
+        pos = BLMSearch._blmname2data[blmname][1]
+        return pos
+    
+    @staticmethod
+    def conv_blmname_2_counter_channel(blmname):
+        """Return the counter channel of a BLM."""
+        chn = BLMSearch._blmname2locs[blmname][2]
```
anacso17, line 43: it seems BLMSearch._blmname2locs is not defined anywere

anacso17:
test is failing because we need to add new class BLMSearch to the list of search interface

siriuspy/siriuspy/search/blm_search.py
xresende: I know. I implemented in blm_search because these BLM are different from other elements: they may not have fixed positions in the ring and their device names (needed for object instantiation of its device class) my change periodically. all related information would be in the search class. this way, when they change location, we do not need to update both search and device classes, but only search (which would contain all the BLM specificities)

another option (maybe cleaner) would be to require PV names to be location-free and the mutable mapping of sensor -> location be in the search class. I did not suggest this approach, which looks better to me, since the begenning because I was not involved in the creation of the IOC. maybe SwC could suggest this and refactor the associated classes.

siriuspy/siriuspy/search/blm_search.py
xresende: same as above

murilobalves (approved):

siriuspy/siriuspy/devices/blm.py
```diff
@@ -0,0 +1,57 @@
...
+from ..search import BLMSearch as _BLMSearch
+from .device import Device as _Device
+
+
+class BLM(_Device):
+    """Beam Loss Monitor Device."""
+
+    class DEVICES:
+        """Devices names."""
+        SENSOR1 = _BLMSearch.conv_blmname_2_devname('SENSOR1')
```
fernandohds564, line 12: Careful here, because if the information regarding BLMs move to the control-system-constants repository in the future, just like the other Search classes infos, this code will communicate with the constants server in import time. We do not do this.

Try to use class properties instead, like [here](https://github.com/lnls-sirius/dev-packages/blob/5d27f2ac8a2bd4fbde0e157d290de111a2e6766f/siriuspy/siriuspy/timesys/csdev.py#L90). This way the code will be robust against this possible change.

siriuspy/siriuspy/search/blm_search.py
xresende: @anacso17 , I removed pvname returning methods from BLM, as we discussed offline.

siriuspy/siriuspy/devices/blm.py
xresende: well pointed. implemented.

siriuspy/siriuspy/devices/blm.py
```diff
@@ -9,17 +10,70 @@ class BLM(_Device):
...
-        SENSOR3 = _BLMSearch.conv_blmname_2_devname('SENSOR3')
-        SENSOR4 = _BLMSearch.conv_blmname_2_devname('SENSOR4')
-        SENSOR5 = _BLMSearch.conv_blmname_2_devname('SENSOR5')
-        SENSOR6 = _BLMSearch.conv_blmname_2_devname('SENSOR6')
-        SENSOR7 = _BLMSearch.conv_blmname_2_devname('SENSOR7')
-        SENSOR8 = _BLMSearch.conv_blmname_2_devname('SENSOR8')
-        ALL = (
-            SENSOR1, SENSOR2, SENSOR3, SENSOR4,
-            SENSOR5, SENSOR6, SENSOR7, SENSOR8)
+        __SENSOR1, __SENSOR2, __SENSOR3, __SENSOR4 = 4 * (None, )
```
fernandohds564, line 13: just as a suggestion, python also allows:
```python3
__SENSOR1 = __SENSOR2 = __SENSOR3 = __SENSOR4 = None
__SENSOR5 = __SENSOR6 = __SENSOR7 = __SENSOR8 = None
```

siriuspy/siriuspy/devices/device.py
```diff
@@ -48,9 +48,10 @@ class Device:
 
     CONNECTION_TIMEOUT = _CONN_TIMEOUT
     GET_TIMEOUT = _GET_TIMEOUT
-    DEVSEP = ':'
     PROPERTIES_DEFAULT = ()
 
+    _DEVSEP = ':'
```
anacso17, line 53: maybe rename to PROPERTY_SEP?

siriuspy/siriuspy/devices/blm.py
xresende: cleaner. I forgot about it. implemented

siriuspy/siriuspy/devices/device.py
xresende: ok. done

anacso17 (approved):

PR #1031, opened 2023-09-28
Update IDSearch

xresende:


anacso17:
TODO list:
- [ ] adapt IDFFConfig to use new IDSearch methods (conv_idname_2_polarization_state)
- [ ] add DELTA52 data in IDSearch

siriuspy/siriuspy/search/id_search.py
```diff
@@ -18,23 +22,77 @@ class IDSearch:
         'CATERETE': 'SI-07SP:ID-APU22',  # titular: VPU19
         'EMA':      'SI-08SB:ID-APU22',  # titular: IVU18
         'MANACA':   'SI-09SA:ID-APU22',  # titular: 2 x APU22
-        'SABIA':    'SI-10SB:ID-EPU50',  # titular: 2 x DELTA52
+        # 'SABIA':    'SI-10SB:ID-EPU50',  # titular: 2 x DELTA52
```
fernandohds564, line 25: why not remove this line?

siriuspy/siriuspy/search/id_search.py
```diff
@@ -18,23 +22,77 @@ class IDSearch:
...
+        'KPARAM_POL_CHANGE',  # [mm]
+        'KPARAM_TOL',  # [mm]
+        'PPARAM_MIN',  # [mm]
+        'PPARAM_MAX',  # [mm]
+        'PPARAM_PARKED',  # [mm]
+        'PPARAM_TOL',  # [mm]
+        )
+
+    _idname2params = {
+        # NOTE: for EPU50 there is a large discrepancy
```
fernandohds564, line 48: why is this note here and not in a more appropriate place?

siriuspy/siriuspy/search/id_search.py
```diff
@@ -18,23 +22,77 @@ class IDSearch:
...
     _idname_2_idff = {
         'SI-06SB:ID-APU22': None,
         'SI-07SP:ID-APU22': None,
         'SI-08SB:ID-APU22': None,
         'SI-09SA:ID-APU22': None,
         'SI-10SB:ID-EPU50': {
-            'polarizations': (
-                'none', 'circularn', 'horizontal', 'circularp', 'vertical',
-                ),
+            'polarizations': tuple(
```
fernandohds564, line 93: why are we still defining properties related to the EPU50?

siriuspy/siriuspy/search/id_search.py
xresende: because this parameter dict contains the primary source of tolerance values for KParam and PParam within our python packages.

siriuspy/siriuspy/search/id_search.py
xresende: EPU50 might be used in the future for beamline commissioning. it is easier (for some of the developers) to have its data structure in an innocuous comment out code than having to find the appropriate commit hash number, one, two years down the road.

siriuspy/siriuspy/search/id_search.py
xresende: see comment below

siriuspy/siriuspy/search/id_search.py
xresende: elaborating on this issue: the appropriate primary source of such parameter/data, in my view, is a table in the control system constant service. the epics base IOC that ARO is developing could get it from csconsts. but this would have to be communicated and implemented (i think this mechanism for epics base IOC would be new). maybe we can talk to ARO and implement this change in the future, if we all agree.

siriuspy/siriuspy/search/id_search.py
fernandohds564: but in this case the EPU will not be used on the SABIA beamline. maybe we don't need to keep this line here.

siriuspy/siriuspy/search/id_search.py
fernandohds564: but the code is not commented, is it?

xresende: the reason I did not was so that we can you the class IDFFConfig (which depends on IDSearch) to get EPU `si_idff` configs int configdb. (there are already some configs saved in the db)

siriuspy/siriuspy/search/id_search.py
xresende: ok, we could remove it.

xresende: done

Gabrielrezende-asc (approved):

fernandohds564 (approved):

anacso17 (approved):

PR #1046, opened 2023-11-27
TIM.FIX: remove FPGAClk-Cte PVs from config.

ericonr:
The value for these PVs is copied from the EVG during startup [1], and should remain constant during operation, otherwise the timing board will lose sync. Furthermore, the AFC IOC will forbid writes to "readonly" PVs in the future [2].

[1] https://github.com/lnls-dig/afc-epics-ioc/blob/b2de36d9e939f064d44d8614aca84811e456dd80/utcaApp/Db/afc_timing.template#L6
[2] https://github.com/lnls-dig/afc-epics-ioc/blob/b2de36d9e939f064d44d8614aca84811e456dd80/iocBoot/iocutca/apply_asg.cmd


anacso17 (approved):

fernandohds564:
only for future reference, a more appropriate prefix for the commit would be: CONFDB.FIX or CONFDB.TIM.FIX

PR #1050, opened 2023-12-06
orbintlk.MNT: de-duplicate FATAL log messages.

ericonr:
It was necessary to add a default argument to
_handle_reliability_failure so it could handle the BPM interlock event without printing a misleading message.

### DIFF
--- a/siriuspy/siriuspy/orbintlk/main.py
+++ b/siriuspy/siriuspy/orbintlk/main.py
@@ -1458,7 +1458,8 @@ class App(_Callback):
 
     def _handle_reliability_failure(self, is_failure=True):
         if is_failure:
-            self._update_log('FATAL:Orbit interlock reliability failure')
+            flag = 'FATAL' if self._state else 'WARN'
+            self._update_log(f'{flag}:Orbit interlock reliability failure')
         if not self._state:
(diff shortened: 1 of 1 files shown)


siriuspy/siriuspy/orbintlk/main.py
```diff
@@ -1461,7 +1456,9 @@ def _check_minsum_requirement(self, monit_sum=None):
         facq_sum = monit_sum * self._get_bpm_rates_factor()
         return _np.all(facq_sum > self._limits['minsum'])
 
-    def _handle_reliability_failure(self):
+    def _handle_reliability_failure(self, is_failure=True):
+        if is_failure:
```
fernandohds564, line 1460: shouldn't this if come after the check for the state?

fernandohds564 (approved):

siriuspy/siriuspy/orbintlk/main.py
fernandohds564: Ok, but, is it a fatal error when the state is not on?

anacso17: Hi guys, I changed to warning when the state is not on

anacso17 (approved):

PR #1057, opened 2023-12-08
IDFF fix and improvement

xresende:


siriuspy/siriuspy/devices/ids.py
```diff
@@ -1278,3 +1278,30 @@ def __init__(self, devname=None, props2init='all', auto_monitor_mon=True):
         # call base class constructor
         super().__init__(
             devname, props2init=props2init, auto_monitor_mon=auto_monitor_mon)
+
+
+class IDFactory():
```
fernandohds564, line 1283: We don't use this strategy of creating a factory in the devices package. See the implementation for SOFB, for instance. There, the class SOFB instantiates the appropriate class, depending on the devname supplied. I think here you could have an `ID` or `InsertionDevice`, that instantiates the appropriate object, instead of an `IDFactory` class, with an `staticmethod` `create` to do so.

fernandohds564 (changes requested):

siriuspy/siriuspy/devices/ids.py
xresende: who are "we" ? I just used this strategy! :smile:

I looked at SOFB now. You propose to use a general device class to instantiate specialized device class objects, instead of using a non-device class for this purpose, right ? Ok, it sounds an alternative possibility (but I see no particular advantage other than using what has been done already). I will call this class `class IDfactory(_Device)`... kidding.

xresende: changed. take a look.

siriuspy/siriuspy/devices/idff.py
```diff
@@ -31,9 +26,9 @@ def __init__(self, devname):
         self._devname = devname  # needed for _create_devices
         self._idffconfig = _IDFFConfig()
 
+        self._pol_mon = _ID.get_idclass(devname).PARAM_PVS.POL_MON
```
anacso17, line 29: I think this line could be after defining `self._devid` and we could use `self._devid.PARAM_PVS.POL_MON` instead of `ID.get_idclass`

fernandohds564 (approved):

siriuspy/siriuspy/devices/ids.py
fernandohds564: Nice, @xresende, thanks. We are all of us, I think. Because all devices  and devices set up to now have a standardized way of instantiation and the factory strategy breaks it, making the library less intuitive.

murilobalves (approved):

siriuspy/siriuspy/devices/idff.py
xresende: you see that I need `pol_mon` in `props2init` before I can create the ID device? But I indeed can simplify things after I changed how I am managing `None` value for `POL_MON` in method `_create_devices`...

xresende: done

anacso17 (approved):

Gabrielrezende-asc (approved):

PR #1060, opened 2023-12-13
Add error log to BLInterlockCtrl device

xresende:
So as to have most diagnostics for failures while opening the beamline.

### DIFF
--- a/siriuspy/siriuspy/devices/intlkctrl.py
+++ b/siriuspy/siriuspy/devices/intlkctrl.py
@@ -149,86 +149,136 @@ class BLInterlockCtrl(_Device):
             devname = self.DEVICES.CAX
         if devname not in self.DEVICES.ALL:
             raise NotImplementedError(devname)
+        self._error_log = ''
         super().__init__(devname, props2init=props2init, **kwargs)
+
+    @property
+    def error_log(self):
+        """Return error log for last check method invoked."""
+        return self._error_log
 
     @property
     def is_hutchA_intlk_search_done(self):
         """."""
-        return self['A:PPS01:SEARCH_OK'] == 1
+        status_ok = self['A:PPS01:SEARCH_OK'] == 1
+        if not status_ok:
+            self._error_log = 'hutch A interlock search not Ok.'
+        return status_ok
 
     @property
     def is_hutchB_intlk_search_done(self):
         """."""
-        return self['B:PPS01:SEARCH_OK'] == 1
+        status_ok = self['B:PPS01:SEARCH_OK'] == 1
+        if not status_ok:
+            self._error_log = 'hutch B interlock search not Ok.'
+        return status_ok
 
     @property
     def is_machine_gamma_enabled(self):
         """."""
-        return self['M:PPS01:HABILITACAO_MAQUINA'] == 1
+        status_ok = self['M:PPS01:HABILITACAO_MAQUINA'] == 1
+        if not status_ok:
+            self._error_log = 'machine gamma not enabled.'
+        return status_ok
 
     @property
     def is_frontend_gamma_shutter_opened(self):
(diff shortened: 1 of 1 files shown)


fernandohds564 (approved):

siriuspy/siriuspy/devices/intlkctrl.py
```diff
@@ -149,86 +149,136 @@ def __init__(self, devname=None, props2init='all', **kwargs):
...
+        """Return error log for last check method invoked."""
+        return self._error_log
+
     @property
     def is_hutchA_intlk_search_done(self):
         """."""
-        return self['A:PPS01:SEARCH_OK'] == 1
+        status_ok = self['A:PPS01:SEARCH_OK'] == 1
+        if not status_ok:
+            self._error_log = 'hutch A interlock search not Ok.'
```
anacso17, line 165: I think error log should return to empty string if command is successful

anacso17 (changes requested):

siriuspy/siriuspy/devices/intlkctrl.py
xresende: Contrary to what I talked to you offline moments agor, ana, I thought about it and I think now that it is more useful for `error_log` to record not the current error state of the BL but the last error observed during the process of opening the BL, regardless of when/what point this happened. For this we should not return empty string in `is_*` methods when they succeed but only when `cmd_beamline_open` specifically succeeds.

xresende: yes. done.

siriuspy/siriuspy/devices/intlkctrl.py
```diff
@@ -268,68 +297,68 @@ def is_hutchB_gatevalves_closed(self):
...
+                self.is_machine_gamma_enabled and \
+                self.is_beamline_eps_ok and \
+                self.is_frontend_shutter_eps_permission_ok and \
+                self.is_hutchA_shutter_eps_permission_ok and \
+                self.is_frontend_gamma_shutter_opened and \
+                self.is_frontend_photon_shutter_opened and \
+                self.is_hutchA_gamma_shutter_opened:
+            return True
+        else:
             return False
```
anacso17, line 361: you could return the if condition here

siriuspy/siriuspy/devices/intlkctrl.py
```diff
@@ -453,60 +481,73 @@ def cmd_hutchA_photon_close(self, timeout=None):
...
                 if _time.time() - t0 > self.TIMEOUT_GATEVALVE:
-                    msg = 'open frontend and hutchA gatevalve timeout reached!'
-                    print(msg)
+                    print(self.error_log)
                     return False
                 _time.sleep(0.5)
 
         # check hutchA shutter permission and open gatevalves for hutchB
         if not self.is_hutchA_shutter_eps_permission_ok:
-            print('open hutchB gatevalves')
```
anacso17, line 493: this removal was on purpose?

siriuspy/siriuspy/devices/intlkctrl.py
xresende: ah, ok. It is True. or False.
done. thanks!

siriuspy/siriuspy/devices/intlkctrl.py
xresende: maybe your comment is deprecated ?

anacso17 (approved):

siriuspy/siriuspy/devices/intlkctrl.py
xresende: yes, I am printing now only if command, which is executed if necessary, fails.

PR #1061, opened 2023-12-15
Add pyproject.toml with Linter/Formatter Configurations

fernandohds564:
This PR tries to define a set of configurations for linters, more specifically `ruff` and `flake8`, and formatters (`ruff`, `black` and `isort`) so that `siriuspy` code becomes more homogeneous. The definitions are located in the new added file `pyproject.toml`. This file serves as source for configuration of several developer tools and also as the configuration of the package itself for installation.

For now I've added the package info contained in `setup.py`, but i left them commented out, so that our installation method remains the same. I also had to add a flag for pip in the `Makefile` so that the installation is not broken. If we remove the comments and the flag from the `Makefile` the installation runs successfully (at least the target `develop-install`), but there are still some edges that need rounding so we can drop the `setup.py` file.

The linting configurations I defined were based on the ones we already followed informally with our old linting methods in vscode. I noticed the `ruff` alone is not complete, so I use both: `ruff` and `flake8`. However, `flake8` does not uses `pyproject.toml` to read its configuration. It seems this is a taboo for them. There was some discussions to add the feature but they were interrupted for being too heated. There are some forks that do implement it, but I decided to stay with the mainstream.

Regarding the formatting, i noticed that `ruff` and `black` are equivalent, with `ruff` being faster. The result I got is not 100% compatible with my aesthetics, but who am I to question them, right? Between us, I particularly don't like the way they break multi-line tuples, arrays, lists or function definitions. I prefer a more compact format. And they prefer double quotes instead of single quotes too...

One thing I noticed is that the formatters almost don't have any option to define how they will format the code. It's all about default. However, `isort`, which only formats the imports, have options. So, what I am doing here is: I run the formatter (`ruff` or `black`) in vscode (the shortcut is `Ctrl+Shit-i`) and after that I fix the imports running `isort`.

NOTE: Due to a bug in the current release of the `isort` pluggin for vscode, the official release will not work for those installations where the source code of our repositories is on the path of the `sirius` mamba environment. In these cases, you need to install the pre-release version of the pluggin that fixes this bug, ok?

As an example of the action of the formatters I formatted the code of `SOFB` in this PR so you guys can compare with the old style I used to follow.

If you agree, I can format other packages in this PR as well.

As an addition, I decided to also keep my vscode configurations for these linters and formatters, so that I can consistently keep the same aesthetics even in projects that don't define them. Follow below the configs I'm currently using in my vscode, in case you guys want to use them:

```json
    "[python]": {
        "editor.formatOnType": false,
        "editor.defaultFormatter": "charliermarsh.ruff",
        // "editor.defaultFormatter": "ms-python.black-formatter",
        "editor.formatOnSave": false,
        "editor.codeActionsOnSave": {
          "source.organizeImports": false,
          "source.fixAll": false
        },
      },

    "flake8.args": [
      "--extend-ignore", "E203",
      "--per-file-ignore", "__init__.py:F401,F821"]

    "ruff.organizeImports": false,
    "ruff.lint.run": "onSave",
    "ruff.run": "onSave",
    "ruff.lint.args": [
      "--select", "W,E,A,B,C90,D,I002,N,F,G,ARG,S,NPY"
      "--ignore",
      "D203,D204,D213,D215,D400,D401,D404,D406,D407,D408,D409,D413,E203,E226",
      "--extend-per-file-ignores", "__init__.py:F821",
      "--extend-per-file-ignores", "__init__.py:F401",
      "--line-length", "79",
      "--preview", "--fix",
    ],
    "ruff.format.args": ["--line-length", "79"],

    "isort.check": true,
    "isort.severity": {"W": "Warning", "E": "Warning"},
    "isort.args": ["--ca", "--cs", "--dt", "-m", "HANGING_INDENT"],

    "black-formatter.args": [
      "--line-length", "79", "--skip-magic-trailing-comma"],
```

Thanks to @VitorSouzaLNLS for helping with the vscode and `pyproject.toml` configuration!


murilobalves (approved):

Gabrielrezende-asc (approved):

VitorSouzaLNLS (approved):

xresende:
I prefer leaving other subpackages to be adapted to ruff after we all had the change to start using this formatter/linter. we can agree to tdo this at the begining of next year, after machine recovery.

xresende (approved):

vellosok75 (approved):

fernandohds564:
@anacso17, what is missing here for us to merge this one?

anacso17 (approved):

PR #1067, opened 2024-01-04
Add imgproc ROIProj PVs

xresende:
- added ImgROITraj PVs with image projections
- defined consts in Const class
- Changed IOC to update PVs that are not of Fit-type even when there is no beam.

ps: Tested!


fernandohds564 (approved):

murilobalves (approved):

xresende:
@anacso17 @RafaelLyra8 we will need a PR in HLA to visualize this intensity projection PV. we can discuss...

PR #1078, opened 2024-03-06
Workaround DCCT reliablility check

xresende:
- discuss this
- should we add a DCCT status on the main monitor window?


xresende:
@anacso17 , I would like to discuss the issue with you. Looking at the currinfo code a posteriori I think this PR is not needed. but I am not sure...

anacso17:
Update: we talked offline and agreed to remove the conditions on ReliableMeas PVs. This will be implemented in a futute PR.

PR #1085, opened 2024-05-20
Add offline data archiver server

xresende:
Extend ClientArchiver classes to allow for accessing archiver server with offline PV data of years previous to 2024.
At this point no archive viewer instance is configured to access this offline data.

Example:

```python 3

import matplotlib.pyplot as plt
from siriuspy.clientarch import PVDataSet, Time

pvnames = ['SI-Glob:AP-CurrInfo:Current-Mon', ]

pvds = PVDataSet(pvnames=pvnames, offline_data=True)

print(pvds.connector.server_url)

pvds.time_start = Time(2021, 5, 1, 12, 0, 0)
pvds.time_stop = Time(2021, 5, 1, 12, 4, 0)

pvds.update()

print(pvds.connector.last_requested_url)

pvd = pvds[pvnames[0]]

plt.plot(pvd.timestamp, pvd.value)
plt.show()

pvds.switch_to_online_data()

```


siriuspy/siriuspy/clientarch/pvarch.py
```diff
@@ -7,15 +7,20 @@
...
 from .client import ClientArchiver as _ClientArchiver
 from .time import Time as _Time, get_time_intervals as _get_time_intervals
 
 
 class _Base:
 
-    def __init__(self, connector=None):
+    DEF_PARALLEL_QUERY_BIN_INTERVAL = 12*60*60  # 12h
+
+    def __init__(self, connector=None, offline_server=False):
```
fernandohds564, line 21: It would be nice to have a mechanism to allow changing the server "in flight", don't you think?

anacso17 (approved):

siriuspy/siriuspy/clientarch/pvarch.py
xresende: but there is. you can use
```Python3

from siriuspy import clientarch as carch

pvname = 'Si-Glob:AP-CurrInfo:Current-Mon'
pvd = carch.PVData(pvdata)
...

# change to offline archiver
pvd.connector.server_url = carch.SERVER_OFFLINE_URL)
```
but I also added helper `switch_to` methods do clientarch class hierarchy to make it easier. take a new look.

fernandohds564 (approved):

siriuspy/siriuspy/clientarch/client.py
```diff
@@ -286,10 +292,19 @@ def getPVDetails(self, pvname, get_request_url=False):
...
         return None if not resp else resp
 
+    def switch_to_online_data(self):
+        """."""
+        self.server_url = _envars.SRVURL_ARCHIVER
+
+    def switch_to_offline_data(self):
+        """."""
+        self.server_url = _envars.SRVURL_ARCHIVER_OFFLINE_DATA
+
```
ericonr, line 302: If you logged in and then switched the data source, the `session` in use will be the wrong one.

I think the `ClientArchiver.switch_to_*_data` methods should also set `self.session = None`. That way, `_handle_request` will automatically let you know you missed a `login` call, if necessary, after calling some of the `switch_to_*_data` methods.

```suggestion
    def switch_to_online_data(self):
        """."""
        self.server_url = _envars.SRVURL_ARCHIVER
        self.session = None

    def switch_to_offline_data(self):
        """."""
        self.server_url = _envars.SRVURL_ARCHIVER_OFFLINE_DATA
        self.session = None

```

ericonr (changes requested):

siriuspy/siriuspy/clientarch/client.py
xresende: you are right, @ericonr . thanks.

xresende: this 'suggested change' ready for commit is very nice! how do I use it ?

ericonr: @xresende when reviewing the code, you can select a range and sometimes (sadly, not always) an "add suggestion" button shows up in the top bar, alongside the formatting buttons. I don't know if you need to be reviewing by commits and not by all changes for that to happen.

image

ericonr (approved):

anacso17 (approved):

siriuspy/siriuspy/clientarch/client.py
xresende: thanks.

PR #1087, opened 2024-05-20
Minor fixes in FOFB IOC

anacso17:
- add new bits in correctors status labels
- initialize new PV values


ericonr (approved):
One issue we noticed during the machine shift was that we had a negative filter gain (due to how the filter was designed). Fixing it in the filter definition by multiplying one of the filters by -1 wasn't an issue, though it had to be done manually, but I wonder if we shouldn't support that case anyway, by changing the lower limit for FOFBAccFilterGain-SP, which is currently 0.

@guilhermerc do you think supporting a negative gain would be interesting, or do you think we can implement that logic in the filter generator script to avoid any manual fixing?

fernandohds564 (approved):

juliacscarvalho (approved):

xresende (approved):

PR #1094, opened 2024-06-25
Code Documentation

juliacscarvalho:
This code documentation uses Sphinx. Feel free to add docstrings. 
To generate the page, run: make html, inside the docs folder, and then open the index.html file inside build folder.

### DIFF
--- a/siriuspy/docs/source/index.rst
+++ b/siriuspy/docs/source/index.rst
@@ -6,10 +6,7 @@
 Welcome to siriuspy's documentation!
 ====================================
 
-.. automodule:: siriuspy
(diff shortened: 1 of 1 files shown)


anacso17:
> @anacso17, what is missing here for merging?

I think we could check if the documentation generated is useful in some way and check if there is some desired docstrings missing to compose the automatic documentation.

anacso17:
> But do you think this is something we should put some effort for now, or is it not a priority for now?

I think it is not a priority for now, but as far as we (also @xresende and @vellosok75 ) have time, it woud be interesting to take a look

PR #1095, opened 2024-06-27
Save PSConfig matrix

juliacscarvalho:
Setting the PSConfig Matrix (previously implemented as FOFBAccFilter) in the High Level FOFB class to be able to save it in the .h5 file.


juliacscarvalho:
I wonder if we could remove the FOFBAccFilter and FOFBAccFilterGain PVs.

anacso17:
> Can we remove the FOFBAccFilter and FOFBAccFilterGain methods?

@juliacscarvalho yes, we should! I forgot that one

siriuspy/siriuspy/devices/fofb.py
```diff
@@ -1575,6 +1576,15 @@ def fofbacc_filter_gain(self):
     def fofbacc_filter_gain(self, value):
         self['FOFBAccFilterGain-SP'] = value
 
+    @property
+    def psconfig_matrix(self):
```
anacso17, line 1580: maybe we could use the same abbreviation used in `respmat`, so this would be `psconfig_mat` ?

anacso17 (approved):

siriuspy/siriuspy/devices/fofb.py
```diff
@@ -1549,31 +1547,13 @@ def fofbacc_decimation_value(self, value):
         self['FOFBAccDecimation-SP'] = value
 
     @property
-    def fofbacc_filter_enumvalue(self):
-        """Accumulator filter enum value."""
-        return self['FOFBAccFilter-Sts']
+    def psconfig_mat(self):
```
murilobalves, line 1550: is the underscore necessary? ```psconfigmat``` seems to be good enough

siriuspy/siriuspy/devices/fofb.py
```diff
@@ -1549,31 +1547,13 @@ def fofbacc_decimation_value(self, value):
...
-    def fofbacc_filter_gain(self):
-        """Accumulator filter gain value."""
-        return self['FOFBAccFilterGain-RB']
-
-    @fofbacc_filter_gain.setter
-    def fofbacc_filter_gain(self, value):
-        self['FOFBAccFilterGain-SP'] = value
+    @psconfig_mat.setter
+    def psconfig_mat(self, value):
+        self['PSConfigMat-SP'] = value
```
murilobalves, line 1556: Perhaps it would be good to add some kind of checking here, for instance check if the shape of the array in ```value``` matches with the readback shape.

murilobalves (changes requested):

fernandohds564 (approved):

anacso17:
@juliacscarvalho , @fernandohds564 , the last commit implies the deploy of these changes must be done for SOFB and FOFB IOCs at the same moment (thanks @fernandohds564 for pointing that)

murilobalves (approved):

xresende (approved):

anacso17 (approved):

PR #1097, opened 2024-07-23
Add BPM polynomial coefficients to as_diagnostics config

anacso17:


siriuspy/siriuspy/clientconfigdb/types/as_diagnostics.py
```diff
@@ -414,6 +414,14 @@ def get_dict():
     [':SwInvGainB-SP', 1.0, 0.0],
     [':SwInvGainC-SP', 1.0, 0.0],
     [':SwInvGainD-SP', 1.0, 0.0],
+    [':GEN_PolyXArrayCoeff-SP', 1.0, 0.0],
```
fernandohds564, line 417: Considering these PVs are arrays, shouldn't you set the default value as an array? maybe the default value could be the array we use today?

fernandohds564 (changes requested):

siriuspy/siriuspy/clientconfigdb/types/as_diagnostics.py
anacso17: done!

fernandohds564 (approved):

PR #1100, opened 2024-08-12
Fix normalizer

xresende:


fernandohds564:
@xresende, isn't this one ready for revision? I think we could merge it and create an image for the IOCs, don't you think? Or you want to do this only after the P7Cav removal?

xresende:
> @xresende, isn't this one ready for revision? I think we could merge it and create an image for the IOCs, don't you think? Or you want to do this only after the P7Cav removal?

I still need to teste IOCs for power supplies other than trims. better wait a little bit.

fernandohds564 (approved):
This solution works, but it doesn't seem very robust.
I'm approving anyway...

murilobalves (approved):

anacso17 (approved):

PR #1126, opened 2024-11-04
Fix bug left in InjCtrl IOC

anacso17:
- The name of the thread used an existing method name
- This is in deploy now


fernandohds564 (changes requested):
@anacso17 , I think there is another bug, because I think the -SP PV of the topup state is not being set by this abort method.

murilobalves (approved):

anacso17:
> @anacso17 , I think there is another bug, because I think the -SP PV of the top-up state is not being set by this abort method.

I think last commit fixes this problem, @fernandohds564

fernandohds564 (approved):

xresende (approved):

PR #1132, opened 2024-11-18
Fix bug in Linac warm up procedure 

anacso17:
Add a wait interval between switching the LLRF LI trigger source and switching the other LI triggers when turning on the LI warm-up period between injections. This seems to solve the problem of the Klystrons loop losing reference at the end of an injection.

Tested during machine study in November 18th.
Update: In deploy since November 18th with https://github.com/lnls-sirius/dev-packages/pull/1126


fernandohds564:
Did this approach worked? I was thinking in maybe changing the actions to turn on and off the triggers state, instead of changing the source of the trigger. I think for all use cases we will want to let the trigger in the continuous mode for some time before injecting the beam... @anacso17 do you remember why we chose this approach of changing the source when we first implemented this?

fernandohds564 (approved):

xresende (approved):

fernandohds564:
@anacso17 I think we can merge this one. Murilo is on vacation and will only come back next year.

PR #1135, opened 2024-11-28
Fix IDFF device use in IDFF IOC class

xresende:
Lately the IDFF device was modified in order to add PVs of the IDFF controller itself, besides those of the ID and those related to correctors. This is convenient for jupyter notebooks that deal with various aspects of the FF system. But the high level FF IOC itself uses the IDFF device. An additional option was added in the IDFF device constructor that allows for choosing when to add the IDFF Ctrl device to the devices set.

### DIFF
--- a/siriuspy/siriuspy/devices/idff.py
+++ b/siriuspy/siriuspy/devices/idff.py
@@ -254,9 +254,10 @@ class IDFF(_DeviceSet):
     class DEVICES(_ID.DEVICES):
         """."""
 
-    def __init__(self, devname):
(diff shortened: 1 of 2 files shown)


fernandohds564 (approved):

xresende:
@anacso17 , this PR has been implicitly tested as it is used by other tested PR.

Gabrielrezende-asc (approved):

anacso17 (approved):

murilobalves (approved):

PR #1146, opened 2025-02-20
Include UpstreamDebugEn-Sel to the AFC TI controlled parameters in Orbit Interlock HLIOC

anacso17:
This helps ensure that upstream debug mode is disabled after any timing machine study.


ericonr (approved):

ericonr (changes requested):
This doesn't work, because `AFCTI_CONFIGS` is locked only for the AFC timings in the critical path, and we need to ensure debugging is disabled for all AFCs

siriuspy/siriuspy/orbintlk/csdev.py
```diff
@@ -107,6 +107,7 @@ class Const(_csdev.Const):
         ('RTMFreqIntgGain-SP', 128),
         ('RTMPhaseNavg-SP', 0),
         ('RTMPhaseDiv-SP', 0),
+        ('UpstreamDebugEn-Sel', 0),
```
ericonr, line 110: I think there should be a comment here explaining our rationale for not locking the value for all crates: we don't want to kill the beam when a timing device outside of the critical path fails, and locking would trigger that, unless we further complicated the locking logic.

We should also update (at least) the orbintlk GUI button that says these PVs refer to the RTM loop.

ericonr (commented):
This was deployed tonight, and behaved correctly.

ericonr (approved):

xresende (approved):

murilobalves (approved):

PR #1147, opened 2025-03-06
Remove dependency on pyepics.

gustavosr8:
This makes the code that uses siriuspy less restricted to some dependencies, such as pyepics, if it is only using ps_search.

### DIFF
--- a/siriuspy/siriuspy/util.py
+++ b/siriuspy/siriuspy/util.py
@@ -9,7 +9,11 @@ import inspect as _inspect
 import subprocess as _sp
 import datetime as _datetime
 
-import epics as _epics
+try:
+    import epics as _epics
+except:
+    _epics = None
+
 
 from mathphys import beam_optics as _beam
 from siriuspy import envars as _envars


siriuspy/requirements.txt
```diff
@@ -2,7 +2,6 @@ aiohttp>=3.7.4
 bottleneck>=1.3.2
 mathphys
 numpy<=1.23
-pyepics>=3.4.0
```
anacso17, line 5: I prefer to keep this requirement

anacso17: So in your environment, you install only those interesting requirements, and do not use this file for that

murilobalves (approved):

ericonr (commented):
> Pyepics uses its own implementation of libCA. If we want to use the upstream lib version and

Pyepics uses its own copy of libCA. If we want to use this module where another version of libCA is already in use and ...

anacso17 (approved):

PR #1155, opened 2025-04-07
Update orbit interlock HLIOC 

anacso17:
- Update LLRF configurations to consider new Input 1 (bit 5) as orbit interlock input
- Add new IA-06RaBPM:TI-EVR to the list of timing redundancy devices

### DIFF
--- a/siriuspy/siriuspy/orbintlk/main.py
+++ b/siriuspy/siriuspy/orbintlk/main.py
@@ -303,7 +303,7 @@ class App(_Callback):
         props_itlk = _ASLLRF.PROPERTIES_INTERLOCK
         devs = [_ASLLRF(devname=name, props2init=props_itlk) for name in names]
         for dev in devs:
-            dev.pv_object('IntlkAll-Mon').auto_monitor = True
+            dev.pv_object('Inp1Intlk-Mon').auto_monitor = True
         return devs
 
     def init_database(self):
@@ -449,12 +449,12 @@ class App(_Callback):
 
     def _handle_lock_llrf(self, dev, init=False):
         dev.wait_for_connection(timeout=self._const.DEF_TIMEOUT)
-        pvo_beamtrip = dev.pv_object('FIMOrbitIntlk-Sts')
+        pvo_beamtrip = dev.pv_object('FIMLLRF1-Sts')
         pvo_manintlk = dev.pv_object('FIMManual-Sts')
         if init:
             pvo_beamtrip.add_callback(_part(
                 self._callback_lock, dev,
-                'FIMOrbitIntlk-Sel', self._llrf_intlk_state))
+                'FIMLLRF1-Sel', self._llrf_intlk_state))
             pvo_manintlk.add_callback(_part(
                 self._callback_lock, dev,
                 'FIMManual-Sel', self._llrf_intlk_state))
@@ -976,8 +976,8 @@ class App(_Callback):
                 name = llrf.system_nickname
                 self._update_log(f'ERR:LLRF-{name} disconnected.')
                 return False
-            llrf.fast_interlock_monitor_orbit = self._llrf_intlk_state
-            llrf.fast_interlock_monitor_manual = self._llrf_intlk_state
+            llrf['FIMLLRF1-Sel'] = self._llrf_intlk_state
+            llrf['FIMManual-Sel'] = self._llrf_intlk_state
         return True
 
     def cmd_config_bpms(self, value):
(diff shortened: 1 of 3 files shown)


siriuspy/siriuspy/orbintlk/main.py
```diff
@@ -1249,8 +1249,8 @@ def _check_configs(self):
         for i, dev in enumerate(self._llrfs):
             if dev.connected:
                 value = _updt_bit(value, 2*i, 0)
-                fim_orbit = dev.fast_interlock_monitor_orbit
-                fim_manual = dev.fast_interlock_monitor_manual
+                fim_orbit = dev['FIMLLRF1-Sts']
```
fernandohds564, line 1252: just out of curiosity, why are you abandoning the use of the python property here @anacso17 ?

siriuspy/siriuspy/orbintlk/main.py
```diff
@@ -1510,10 +1512,10 @@ def _handle_reliability_failure(self, is_failure=True):
         self._update_log('FATAL:sending soft interlock to LLRF.')
         # sending interlock for all LLRFs systems, then wait
         for llrf in self._llrfs:
-            llrf.interlock_manual = 1
+            llrf['IntlkManual-Sel'] = 1
         _time.sleep(1)
         for llrf in self._llrfs:
-            llrf.interlock_manual = 0
+            llrf['IntlkManual-Sel'] = 0
```
fernandohds564, line 1518: same here.

siriuspy/siriuspy/orbintlk/main.py
anacso17: Hi @fernandohds564 , this was a first step for testing, in a next commit I would replace the hard-coded strings by attributes depending on the LLRF, this would make it possible to have different inputs for each LLRF (as we needed the last months)

siriuspy/siriuspy/orbintlk/main.py
anacso17: In fact here we could use the property

siriuspy/siriuspy/orbintlk/main.py
fernandohds564: thanks for the explanation!

ericonr (approved):
Changes suggested by fernando haven't been pushed yet. Besides that, LGTM

siriuspy/siriuspy/orbintlk/main.py
```diff
@@ -1510,10 +1523,10 @@ def _handle_reliability_failure(self, is_failure=True):
...
```
ericonr, line 1529: Not using the property

siriuspy/siriuspy/orbintlk/main.py
```diff
@@ -1249,8 +1260,8 @@ def _check_configs(self):
         for i, dev in enumerate(self._llrfs):
             if dev.connected:
                 value = _updt_bit(value, 2*i, 0)
-                fim_orbit = dev.fast_interlock_monitor_orbit
-                fim_manual = dev.fast_interlock_monitor_manual
+                fim_orbit = dev['FIMLLRF1-Sts']
+                fim_manual = dev['FIMManual-Sts']
```
ericonr, line 1264: Shouldn't we keep using the property for manual, since that's not going to change depending on the LLRF? It's not a physical connection, after all.

siriuspy/siriuspy/orbintlk/main.py
anacso17: done

siriuspy/siriuspy/orbintlk/main.py
anacso17: Ok, done

anacso17:
@fernandohds564 , @ericonr , I changed, where was possible, to use properties. I talked to @felipekojihoshino and it seems to be definitive for both LLRF the FIMLLRF1 as orbit interlock input. 

(I have changed to use only PV names directly also because it was annoying me to have two different ways to access the same PV (using the PV name when I need the PV object and using the property when I needed only reading or writing the value), but I see it is interesting to use properties when it is possible, to avoid hard-coded strings. Unfortunately it is not possible to avoid this when we need the PV object itself)

ericonr:
Nit: `orbintlk.FIX: fix timing redundancy check` commit body could be wrapped to 72 columns

And I was wondering if maybe the Fout checking shouldn't (eventually) happen in the high level trigger?

anacso17:
> Nit: orbintlk.FIX: fix timing redundancy check commit body could be wrapped to 72 columns

done!

PR #1157, opened 2025-04-10
Add method to set TbT mask to FamBPMs

vellosok75:


siriuspy/siriuspy/devices/bpm_fam.py
```diff
@@ -221,6 +221,46 @@ def set_attenuation(self, value=RFFEATT_MAX, timeout=TIMEOUT):
             _log.info(mstr)
         return okall
 
+    def set_tbt_mask(self, mask_begin, mask_end, timeout=TIMEOUT):
+        """."""
+        if isinstance(mask_begin, int):
+            mask_begin = _np.ones(len(self.devices)) * mask_begin
+        if isinstance(mask_end, int):
+            mask_end = _np.ones(len(self.devices)) * mask_end
```
fernandohds564, line 229:
```suggestion
        if isinstance(mask_begin, (int, float)):
            mask_begin = _np.full(len(self.devices), mask_begin)
        if isinstance(mask_end, (int, float)):
            mask_end = _np.full(len(self.devices), mask_end)

        mask_begin = _np.array(mask_begin)
        mask_end = _np.array(mask_end)
        
        if {mask_begin.size, mask_end.size} != {len(self.devices)}:
            return False
```

siriuspy/siriuspy/devices/bpm_fam.py
```diff
@@ -221,6 +221,46 @@ def set_attenuation(self, value=RFFEATT_MAX, timeout=TIMEOUT):
             _log.info(mstr)
         return okall
 
+    def set_tbt_mask(self, mask_begin, mask_end, timeout=TIMEOUT):
```
fernandohds564, line 224: I think we should have an input variable to decide whether we want to enable or disable the mask and also mask_begin and mask_end should have a default value of None, which means it would not set that property.

In this case, the suggestion bellow would not be valid and I think a refactoring of this method would be advised.

siriuspy/siriuspy/devices/bpm_fam.py
```diff
@@ -243,6 +244,13 @@ def _to_array(val, name):
         mask_begin = _to_array(mask_begin, "mask_begin")
         mask_end = _to_array(mask_end, "mask_end")
 
+        total_samples = mask_begin if mask_begin is not None else 0
+        total_samples += mask_end if mask_end is not None else 0
+        if _np.any(total_samples >= 382):
```
fernandohds564, line 249: hard-coded number.

fernandohds564 (approved):

PR #1170, opened 2025-06-05
Add auto_monitor option to SOFB device

xresende:


xresende:
> @xresende, I left comments with suggestions below. I think it would be helpful if you configure ruff formatter in your vscode.

I thought I already had it! 

image

murilobalves (approved):

xresende:
> Maybe you do have it installed, but I think it is not configured properly. Either way, did you run the formatter on these changes?



> Maybe you do have it installed, but I think it is not configured properly. Either way, did you run the formatter on these changes?

@fernandohds564 , in a few days time, after I move to the same office as you, we can sit together and you can help me check my ruff configuration.

fernandohds564 (approved):

xresende:
@vellosok75 ?

vellosok75 (approved):

PR #1171, opened 2025-06-06
Modernize Install

fernandohds564:
See discussion in lnls-sirius/scripts#180

PR #1177, opened 2025-07-13
Improve update of MTurn and SP orbits in SOFB

fernandohds564:
This change will decrease archiver load, since triggered orbits will only update when all BPMs have updated.

to be merged after #1171


anacso17:
It seems to be based on https://github.com/lnls-sirius/dev-packages/pull/1171, it needs to wait the review for that one.

fernandohds564:
@anacso17, I'll close this one and open another that doesn't depend on #1171, because I think we could merge these changes asap.

PR #1178, opened 2025-07-14
Improve update of MTurn and SP orbits in SOFB

fernandohds564:
This change will decrease archiver load, since triggered orbits will only update when all BPMs have updated.

I also removed @murilobalves from the CODEOWNERS file so that he is not included automatically, because, differently from what I thought, we cannot remove him from the reviewers list before creating the PR, which means he would continue to receive notifications for every PR.

### DIFF
--- a/CODEOWNERS
+++ b/CODEOWNERS
@@ -1,2 +1,2 @@
 # --- default owners of the repo
-* @anacso17 @fernandohds564 @murilobalves @xresende 
+* @anacso17 @fernandohds564 @xresende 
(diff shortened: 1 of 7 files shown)


fernandohds564:
> @fernandohds564 approving this but maybe it is worth it to study the possibility to use devices for the classes that handle connections to PVs.

@anacso17, I had thought about this several times already. The problem is that our `Device` subclasses does not have a key functionality I need here, which is the `_config_ok_vals`, where I keep the configurations I want for each PV. If I were to use the `Device` objects, I would have to implement this another way, which would consume more time and would probably not be as elegant as the current implementation, so I never tried it. 

But I invite you to make the necessary changes and submit a PR!
We can test it together in some machine study.

xresende (approved):

PR #1197, opened 2025-09-19
Fix wfmsize bug

xresende:


xresende:
> hi @xresende , the siriushla package depends on the waveform variables, so I think we need to keep compatibility, not changing the variable names. I think it is not convenient to break this compatibility.

I created a PR in HLA

siriuspy/siriuspy/pwrsupply/psctrl/pscontroller.py
```diff
@@ -20,6 +20,7 @@ class PSController:
         'WfmOffsetKL-SP', 'WfmOffsetKL-RB',
         'WfmOffsetSL-SP', 'WfmOffsetSL-RB',
         'WfmOffsetEnergy-SP', 'WfmOffsetEnergy-RB',
+        'Log-Mon',  # This PV is related to the IOC, not PSController
```
anacso17, line 23: Is this PV defined in database? I think it is only defined in other PR

PR #1199, opened 2025-09-24
Fix HLFOFB IOC initialization

anacso17:
This PR fixes the strange behavior of having to set response and PS configuration matrices every time we restart HLFOFB IOC.


ericonr (approved):
If you could copy your PR description to your commit message!

> This commit fixes the strange behavior of having to set response and PS configuration matrices every time we restart HLFOFB IOC.

PR #1230, opened 2026-03-24
No(Only One) Thread in ClientArch

fernandohds564:
When we updated the `clientarch` subpackage to use asynchronous requests to the server under the hood, we implemented it so that all `async` methods run in a separate thread. This was initially thought to be necessary to avoid conflicts in environments where an event loop is already running, such as Jupyter notebooks.

However, this approach introduces several issues. Exceptions raised in that separate thread cannot be caught by the end user, and errors do not interrupt the execution flow as expected. In addition, there are cases where the thread silently dies, causing user calls to hang indefinitely.

~~It turns out there is a better way to run `async` methods under the hood without exposing `async` in the public API. This PR attempts to implement that approach. Most of the implementation was done by GitHub Copilot, with a few manual adjustments on my part. One of those changes is the introduction of a dependency on the `nest_asyncio` package.~~

Since the `nest_asyncio` package was archived, we implemented again a solution based on running a different event loop in a new thread. However, this thread is long lived and all tasks are submitted using `asyncio.run_coroutine_threadsafe` function. This guarantees that exceptions will be passed to the end user and that there will be no infinite hangs.

This last version of the code was done with the help of Google AI Mode.

This PR also solves a bug related on how to properly use `aiohttp`. Now the `json` parsing of the response happens within a `async with` for the get resquest. Thanks to @anacso17 for helping with this debugging.

The bug fix above, made it impossible to keep the old method to check for connection in the `ClientArchiver.connected` property, since we need to convert the response, either to `json` or to `text` within the `async with` context manager. For this reason, the option `return_json` was removed. Now all requests will return data in `json` format, falling back to pure text data when `json` conversion fails, which is the case of the `ClientArchiver.connected` method.

This PR also fixes the `ClientArchiver.login()` method that was not working in computers where the server is defined by its IP address, instead of its hostname (control room computers).

### This version of the code was tested in our conda environments and at the control room.


siriuspy/requirements.txt
```diff
@@ -1,6 +1,7 @@
 aiohttp>=3.7.4
 bottleneck>=1.3.2
 mathphys
+nest_asyncio
```
anacso17: This project was discontinued by the author (archived https://github.com/erdewit/nest_asyncio), so I think it is not a good idea to include it as a dependency

anacso17 (changes requested):

siriuspy/requirements.txt
fernandohds564: right @anacso17. It seems the project was discontinued due to unforeseen circumstances. Besides, it seems that the topic of nested tasks was intensely discussed among the Python core team and decided not to be a good practice after all.

I removed the dependency and solved the problem with a different aproach, falling back to the new event loop in a different thread, but without the downsides we had with the previous implementation. Please, take a look.

anacso17 (approved):

xresende (approved):

fernandohds564:
Guys, @anacso17 @vellosok75 @afonsoln @VitorSouzaLNLS , could you take a look at this PR, please?

fernandohds564:
Guys, @anacso17 @vellosok75 @afonsoln @VitorSouzaLNLS , could you take a look at this PR, please?

anacso17 (changes requested):
Hi @fernandohds564, as we discussed offline, running Machine Reports using this branch gave me very strange results that I don't understand (whether it's related to the payload settings or something else). Because of this, I'm blocking the merge here until we can understand and correct this behavior.

fernandohds564:
Talking offline to @anacso17, we found out the problem is actually related to the master branch. So I'll request another review from @anacso17.

anacso17 (approved):
As we talked offline, we discovered that sometimes the master version of the `ClientArchiver` class returns "None" values even for PVs that are being archived in the period requested. For this new version, this problem seems to be solved, and because of that some reports were returning very different values. We understood the source of the difference, and I will create soon a PR to fix the Machine Reports.

anacso17:
Just for completion, PR for Machine Reports fixes: https://github.com/lnls-sirius/dev-packages/pull/1247

PR #1231, opened 2026-03-27
Improve API of clientarch

fernandohds564:
To be merged after #1230.

This PR changes the interface of the classes `PVData`, `PVDataSet` and `ClientArchiver` from `siriuspy.clientarch` package.

The main modifications are:
 - renaming of all methods of `ClientArchiver` to snake_case, the python standard;
 -  addition of new funcionalities in `ClientArchiver`:
     - `login` now takes the password as an optional argument. If not provided, it will prompt the user for the password, in a discrete way. The old interface required the user to write its password in scripts and or ipython terminals, which would result in sensitive data being saved in files and logged in histories, etc...
     - `get_data` now supports all types of processing the archiver interface provides, not just the `'mean'`. It takes two parameters, namely `proc_type_param1` and `proc_type_param2` to control the parameters of the chosen statistics. Some statistics may use one or none of them;
     - `get_report` was created and supports all types of reports provided by the archiver interface;
     - `get_detailed_appliance_metrics` was created and provides the results of the `'getApplianceMetricsForAppliance'` call to archiver;
     - `get_process_metrics_for_appliance` was created and provides the results of the `'getProcessMetricsDataForAppliance'`;
     - the property `timeout` was renamed to `query_timeout` to better reflect its meaning: the timeout of each query transaction, instead of the total timeout of all queries.
     - the property `parallel_query_bin_interval` was renamed to `query_split_interval`. Now it will always be taken into consideration when creating queries urls, for splitting large time intervals into smaller ones;
     - the option `parallel` of `get_data` was removed. we can achieve similar results by setting `query_split_interval` to `0`;
     - the property `query_max_concurrency` was created to limit the maximum number of concurrent queries that will be requested from the server. It is an additional control for optimizing the retrieval time.
 - changes/improvements of `PVData` API:
     - implementation of `__str__` method to show a nice visualization of the class properties
     - removal of `timestamp_start` and `timestamp_stop`. Now only `time_start` and `time_stop` exists. Their getter return a `clientarch.Time` but their setters accept anything that can be converted to `Time` object, including timestamps;
     - the properties `processing_type`, `processing_type_param1`, `processing_type_param2` and `query_split_interval` were added to control how `self.connector.get_data` will be called inside the `self.update` method. This means the `self.update` method does not take any arguments beside a `query_timeout`, that overwrites `self.connector.query_timeout` if passed;
     - the properties `query_max_concurrency` and `query_timeout` were added to ease setting of the `self.connector` respective properties;
 - Changes/improvements of `PVDataSet` API:
     - implementation of `__str__` method to show a nice visualization of the class properties
     - removal of `timestamp_start` and `timestamp_stop`. Now only `time_start` and `time_stop` exists. Their getter return a `clientarch.Time` if all internal `PVData` objects have equal `start_time|stop_time` or a list of them, one for each `PVData` object. Their setters accept anything that can be converted to `Time` object, including timestamps. It also accepts a list of `Time` objects, referring to the desired `start_time|stop_time` of each `PVData`.
     - the properties `processing_type`, `processing_type_param1`, `processing_type_param2` and `query_split_interval` were added to control how `self.connector.get_data` will be called inside the `self.update` method. Their getters return a single value or a list of values depending if the associated `PVData`'s are identical or not, and the setter also accepts a list of values, one for each `PVData`. This means the `self.update` method does not take any arguments beside a `query_timeout`, that overwrites `self.connector.query_timeout` if passed;
     - the properties `query_max_concurrency` and `query_timeout` were added to ease setting of the `self.connector` respective properties;

**This PR was testes in the control room computers, using the script `gensumm_machreport.py` and other manual tests.**


fernandohds564:
Guys, @xresende  @anacso17 @vellosok75 @afonsoln @VitorSouzaLNLS , could you take a look at this PR, please?

xresende:
@fernandohds564 ,

I am trying the PR in a notebook...
intuitively I tried:

image

and then I realized this is not how the argument to the new method should be passed. I think it is useful to have class constants from which we can drawn the constant strings through code completion.

fernandohds564:
> @fernandohds564 ,
> 
> I am trying the PR in a notebook... intuitively I tried:
> 
> image
> and then I realized this is not how the argument to the new method should be passed. I think it is useful to have class constants from which we can drawn the constant strings through code completion.

yes, you are right, the method should accept both options, the name and the value of the namedtuple.

xresende (approved):

anacso17 (changes requested):
Hi @fernandohds564, unfortunately I had the same problems I mentioned on this https://github.com/lnls-sirius/dev-packages/pull/1230#pullrequestreview-4358823168, that's why I am blocking merge here too.

fernandohds564:
Talking offline to @anacso17, we found out the problem is actually related to the master branch. So I'll request another review from @anacso17.

anacso17 (approved):
Same as https://github.com/lnls-sirius/dev-packages/pull/1230#pullrequestreview-4405186356

PR #1235, opened 2026-04-28
Add device BO tune

xresende:
needs https://github.com/lnls-sirius/dev-packages/pull/1236

### DIFF
--- a/siriuspy/siriuspy/devices/tune.py
+++ b/siriuspy/siriuspy/devices/tune.py
@@ -144,6 +144,54 @@ class SITuneProc(_Device):
         return self['Trace-Mon']
 
 
+class BOTuneProc(_Device):
+    """Tune Proc device."""
+
+    class DEVICES:
+        """Devices names."""
+
+        H = 'BO-Glob:DI-TuneProc-H'
+        V = 'BO-Glob:DI-TuneProc-V'
+        ALL = (H, V)
+
+    PROPERTIES_DEFAULT = (
+        'SpecArray-Mon',
+        'SwePts-SP',
+        'SwePts-RB',
+        'FrameCount-Mon',
+    )
+
+    def __init__(self, devname, props2init='all'):
+        """Init."""
+        # check if device exists
+        if devname not in BOTuneProc.DEVICES.ALL:
+            raise NotImplementedError(devname)
+
+        # call base class constructor
+        super().__init__(devname, props2init=props2init)
+
+    @property
+    def sweep_nrpts(self):
+        """."""
+        return int(self['SwePts-RB'])
+
+    @sweep_nrpts.setter
+    def sweep_nrpts(self, value):
+        """."""
+        self['SwePts-SP'] = value
+
+    @property
+    def frame_count(self):
+        """."""
+        return self['FrameCount-Mon']
+
+    @property
+    def tune_spect_array(self):
+        """Tune spect array."""
+        spect = self['SpecArray-Mon']
+        return spect
+
+
 class Tune(_DeviceSet):
     """Tune device."""
 
--- a/siriuspy/siriuspy/devices/tune.py
+++ b/siriuspy/siriuspy/devices/tune.py
@@ -432,13 +432,12 @@ class Tune(_DeviceSet):
 
     def _get_spect(self, dev_proc):
         """."""
-        if self.sector == 'BO':
-            nr_pulses = self.trig_nr_pulses
-            sweep_nrpts = dev_proc.sweep_nrpts
-            data = dev_proc.tune_spect_array
-            data = data[:sweep_nrpts*nr_pulses]
-            spect = data.reshape(nr_pulses, -1)
-            return spect
+        nr_pulses = self.trig_nr_pulses
+        sweep_nrpts = dev_proc.sweep_nrpts
+        data = dev_proc.tune_spect_array
+        data = data[:sweep_nrpts*nr_pulses]
+        spect = data.reshape(nr_pulses, -1)
+        return spect
 
 
 class TuneCorr(_Device):
--- a/siriuspy/siriuspy/devices/tune.py
+++ b/siriuspy/siriuspy/devices/tune.py
@@ -1,6 +1,7 @@
 """Tune devices."""
 
 from .device import Device as _Device, DeviceSet as _DeviceSet
+from .timing import Trigger as _Trigger
 
 
 class TuneFrac(_Device):
@@ -54,7 +55,7 @@ class TuneFrac(_Device):
     def tune(self):
         """Tune Frac."""
         if self.sector == 'SI':
-        return self['TuneFrac-Mon']
+            return self['TuneFrac-Mon']
 
     @property
     def span(self):
@@ -199,7 +200,8 @@ class Tune(_DeviceSet):
         """Devices names."""
 
         SI = 'SI-Glob:DI-Tune'
-        ALL = (SI,)
+        BO = 'BO-Glob:DI-Tune'
+        ALL = (SI, BO)
 
     def __init__(self, devname=None, props2init='all'):
         """Init."""
@@ -208,6 +210,7 @@ class Tune(_DeviceSet):
             devname = Tune.DEVICES.SI
         if devname not in Tune.DEVICES.ALL:
             raise NotImplementedError(devname)
+        self._sector = 'SI' if devname.startswith('SI') else 'BO'
 
         isall = isinstance(props2init, str) and props2init.lower() == 'all'
         if not isall and props2init:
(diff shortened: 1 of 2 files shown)


siriuspy/siriuspy/devices/tune.py
```diff
@@ -225,15 +225,35 @@ def __init__(self, devname=None, props2init='all'):
...
+    @property
+    def dev_tune_proc_v(self):
+        """."""
+        return self.devices[3]
+
     @property
     def tunex(self):
         """Tune Frac X."""
-        return self.devices[0].tune
+            return self.dev_tune_frac_h.tune
```
xresende, line 298: ops. still learning how to commit sections of a file. this line makes sense with the next commit

siriuspy/siriuspy/devices/tune.py
```diff
@@ -225,15 +225,35 @@ def __init__(self, devname=None, props2init='all'):
...
 
     @property
     def tuney(self):
         """Tune Frac Y."""
-        return self.devices[1].tune
+            return self.dev_tune_frac_v.tune
```
xresende, line 304: same here

siriuspy/siriuspy/devices/tune.py
```diff
@@ -54,7 +55,7 @@ def sector(self):
     def tune(self):
         """Tune Frac."""
         if self.sector == 'SI':
-        return self['TuneFrac-Mon']
+            return self['TuneFrac-Mon']
```
xresende, line 59: a fix

siriuspy/siriuspy/devices/tune.py
```diff
@@ -26,17 +30,32 @@ class DEVICES:
...
+        if self.sector == 'SI':
+            self.PROPERTIES_DEFAULT = TuneFrac.PROPERTIES_DEFAULT_SI
+        else:
+            self.PROPERTIES_DEFAULT = TuneFrac.PROPERTIES_DEFAULT_BO
         super().__init__(devname, props2init=props2init)
 
+    @property
+    def sector(self):
+        """."""
+        return self._sector
```
fernandohds564: try to use the properties of `SiriusPVName` to get the sector, instead of creating this hidden attribute:
```suggestion
        return self.devname.sec
```

siriuspy/siriuspy/devices/tune.py
```diff
@@ -26,17 +30,32 @@ class DEVICES:
...
+    PROPERTIES_DEFAULT_SI = PROPERTIES_DEFAULT + (
+        'TuneFrac-Mon',
+    )
+    PROPERTIES_DEFAULT_BO = PROPERTIES_DEFAULT
 
     def __init__(self, devname, props2init='all'):
         """Init."""
         if devname not in TuneFrac.DEVICES.ALL:
             raise NotImplementedError(devname)
+        self._sector = 'SI' if devname.startswith('SI') else 'BO'
```
fernandohds564: delete this line. See suggestion below.
```suggestion
```

siriuspy/siriuspy/devices/tune.py
```diff
@@ -127,32 +210,134 @@ def __init__(self, devname=None, props2init='all'):
             devname = Tune.DEVICES.SI
         if devname not in Tune.DEVICES.ALL:
             raise NotImplementedError(devname)
+        self._sector = 'SI' if devname.startswith('SI') else 'BO'
```
fernandohds564: remove this line.
```suggestion
```

siriuspy/siriuspy/devices/tune.py
```diff
@@ -127,32 +210,134 @@ def __init__(self, devname=None, props2init='all'):
...
+                trig
+            )
 
         # call base class constructor
         super().__init__(devices, devname=devname)
 
+    @property
+    def sector(self):
+        """."""
+        return self._sector
```
fernandohds564: same as suggestion above
```suggestion
        return self._devname.sec
```

siriuspy/siriuspy/devices/tune.py
```diff
@@ -111,14 +145,63 @@ def tune_wfm(self):
...
+
+    @property
+    def frame_count(self):
+        """."""
+        return self['FrameCount-Mon']
+
+    @property
+    def tune_spect_array(self):
+        """Tune spect array."""
+        spect = self['SpecArray-Mon']
```
fernandohds564, line 193: maybe we should apply a reshape here:
```suggestion
        spect = self['SpecArray-Mon']
        swnpts = self.freq_sweep_nrpts
        siz = (spect.size // swnpts) * swnpts
        spect = spect[:siz].reshape(-1, swnpts)
```

siriuspy/siriuspy/devices/tune.py
```diff
@@ -229,21 +414,32 @@ def enabley(self):
...
+    def _get_spect(self, dev_proc):
+        """."""
+        nr_pulses = self.trig_nr_pulses
+        freq_sweep_nrpts = dev_proc.freq_sweep_nrpts
+        size = freq_sweep_nrpts * nr_pulses
+        data = dev_proc.tune_spect_array
+        data = data[:size]
+        spect = data.reshape(nr_pulses, -1)
+        spect = spect[::-1, :]
+        return spect
```
fernandohds564: I think the better option is to return `frame_count` measurements, instead of `trig_nr_pulses`.
This way, and assuming the suggestion above was accepted, we could change here to:
```suggestion
        spect = dev_proc.tune_spect_array[:dev_proc.frame_count]
        return spect[::-1]
```

siriuspy/siriuspy/devices/tune.py
xresende: good. done

siriuspy/siriuspy/devices/tune.py
xresende: done

siriuspy/siriuspy/devices/tune.py
xresende: done

siriuspy/siriuspy/devices/tune.py
xresende: done

siriuspy/siriuspy/devices/tune.py
xresende: good idea

siriuspy/siriuspy/devices/tune.py
xresende: right. merged

siriuspy/siriuspy/devices/tune.py
```diff
@@ -69,37 +91,52 @@ def center_frequency(self, value):
...
+        """Return aquisition state."""
+        return self['SpecAnaGetSpec-Sts']
+
+    @property
+    def excitation_enabled(self):
+        """Return excitation state."""
         return self['Enbl-Sts']
 
-    def cmd_enable(self, timeout=DEF_TIMEOUT):
+    def cmd_acquisition_enable(self, timeout=DEF_TIMEOUT):
```
fernandohds564, line 101: I think it is better if the name is closer to the action: `cmd_enable_acquisition`, don't you think?

siriuspy/siriuspy/devices/tune.py
```diff
@@ -127,32 +216,133 @@ def __init__(self, devname=None, props2init='all'):
...
-        devices = (tune_frac_h, tune_frac_v, tune_proc_h, tune_proc_v)
+        if devname.sec == _SI.sector:
+            tune_frac_h = TuneFrac(
+                TuneFrac.DEVICES.SI_H, props2init=props2init)
+            tune_frac_v = TuneFrac(
+                TuneFrac.DEVICES.SI_V, props2init=props2init)
+            tune_proc_h = SITuneProc(
+                SITuneProc.DEVICES.H, props2init=props2init)
+            tune_proc_v = SITuneProc(
+                SITuneProc.DEVICES.V, props2init=props2init)
```
fernandohds564: do you really need to break lines here? If you do, the parenthesis should be on another line? Didn't your formatter catch this?

siriuspy/siriuspy/devices/tune.py
```diff
@@ -229,25 +419,46 @@ def enabley(self):
         """Tune Y enable status."""
         return self.devices[1].enable
 
-    def cmd_enablex(self, timeout=TuneFrac.DEF_TIMEOUT):
-        """Enable tune X."""
-        return self.devices[0].cmd_enable(timeout=timeout)
+    def cmd_acquisition_enablex(self, timeout=TuneFrac.DEF_TIMEOUT):
```
fernandohds564, line 423: same comment for these names.

siriuspy/siriuspy/devices/tune.py
xresende: i like method names to be grouped first by category rather than by action, i.e.:

`cmd -> {acquisition|excitation|...} -> {enable|disable}`

to me this ordering feels more natural and scales better as the number of subsystems and actions grows. For example, when browsing code completion results or grepping for methods, having:

* `cmd_acquisition_enable`
* `cmd_acquisition_disable`
* `cmd_acquisition_reset`

keeps all acquisition-related commands grouped together. The same naturally happens for excitation, triggering, configuration...

With the action-first style:

* `cmd_enable_acquisition`
* `cmd_enable_excitation`
* `cmd_disable_acquisition`

methods become grouped primarily by verb (`enable`, `disable`, ...), which I personally find less useful when working within a specific subsystem. i also think the subsystem-first naming scales better once operations become more than simple binary enable/disable actions. For example:

* `cmd_acquisition_configure`
* `cmd_acquisition_trigger`
* `cmd_acquisition_abort`

still fit naturally within the same hierarchy.

A `grep -R cmd_` in `dev-packages` shows both naming styles in use. Most of the code I implemented naturally followed this ordering, while SOFB commands appear to use both conventions. But the majority of the existing codebase seems to follow your suggested style.

siriuspy/siriuspy/devices/tune.py
xresende: done

siriuspy/siriuspy/devices/tune.py
xresende: see argument above

siriuspy/siriuspy/devices/tune.py
xresende: so what should we do ? do you still think we should adapt this PR to your cmd method naming style or should we standardize all cmd_ methods in a separete PR, @fernandohds564 ?

fernandohds564: I think its ok. My suggestions was mainly due to the lack of specification in the previous naming. This standard is useful.

fernandohds564 (approved):

PR #1236, opened 2026-05-05
ENH: optics consts classes

xresende:


siriuspy/siriuspy/optics/constants.py
```diff
@@ -1,25 +1,91 @@
...
+
+
+class _ClassProperty:
+    """."""
+
+    def __init__(self, func):
+        self.func = func
+
+    def __get__(self, obj, cls):
+        return self.func(cls)
```
fernandohds564: We already have an implementation of a classproperty at `siriuspy.utils.ClassProperty`. Why re-implementing it here?

siriuspy/siriuspy/optics/constants.py
```diff
@@ -1,25 +1,91 @@
...
+    @_ClassProperty
+    def beam_rigidity(cls):
+        """Beam ridigity."""
+        brho = _bopts.beam_rigidity(energy=cls.beam_energy)
+        return brho
+
+    @_ClassProperty
+    def velocity(cls):
+        """Beam velocity [m/s]."""
+        _, velocity, *_ = cls.beam_rigidity
```
fernandohds564: Is this correct? According to the code above, `cls.beam_rigidity` only returns `brho`, right?

siriuspy/siriuspy/optics/constants.py
```diff
@@ -1,25 +1,91 @@
...
+        return velocity
+
+
+class TLine(Accelerator):
+    """Transport line."""
+
 
 class Ring(Accelerator):
     """Ring."""
```
fernandohds564, line 27: I think you should define the length attribute here.

fernandohds564 (commented):
I left a few comments in the code.

siriuspy/siriuspy/optics/constants.py
xresende: ops. did not realize it was already implemented. done.

siriuspy/siriuspy/optics/constants.py
xresende: it is implemented as intended. 
cls.beam_rigidity returns the same tuple that mathphys.beam_optics.beam_rigidity returns...

siriuspy/siriuspy/optics/constants.py
xresende: It is already defined in the base class `Accelerator`

xresende: ah, you mean the correct constant values? they are defined under TB, TS, BO and SI...

fernandohds564: yes, but since this is a common attribute of all derived classes, I think it should be defined here as well to define the interface of the class.

siriuspy/siriuspy/optics/constants.py
fernandohds564: yes, but do you think this name is intuitive for what is being returned? I expected beam_rigidity to return the magnetic rigidity of the beam, not a tuple will several parameters.

xresende: I am just using what is implemented in mathphys and siriuspy.utils, @fernandohds564. are you suggesting we should change everything now? If this the case I suggest doing this in separate PRs

xresende: ok, discussed offline. I will delete beam_rigidity, leave velocity property. in the future we should change beam_rigidity in mathphys and siriuspy.utils to return a beam class object with beam parameters as attributes/methods.

xresende: > ok, discussed offline. I will delete beam_rigidity, leave velocity property. in the future we should change beam_rigidity in mathphys and siriuspy.utils to return a beam class object with beam parameters as attributes/methods.

done.

fernandohds564 (approved):

anacso17 (approved):

anacso17 (approved):

PR #1237, opened 2026-05-06
Use new optics consts classes

xresende:
needs https://github.com/lnls-sirius/dev-packages/pull/1236


siriuspy/siriuspy/sofb/csdev.py
```diff
@@ -135,7 +141,7 @@ class SOFBTLines(ConstTLines):
 
     def __init__(self, acc):
         """Init1 method."""
-        self.acc = acc.upper()
+        self.acc = acc
```
fernandohds564, line 140: I don't understand. How can you be sure that `acc` is one of the classes you defined? I think you have to parse this argument here.

siriuspy/siriuspy/sofb/csdev.py
```diff
@@ -204,8 +210,8 @@ def __init__(self, acc):
         self.evt_acq_name = "Linac"
         self.matrix_size = self.nr_corrs * (2 * self.nr_bpms)
         self.nr_svals = min(self.nr_corrs, 2 * self.nr_bpms)
-        self.circum = 21.2477 if self.acc == "TB" else 26.8933  # in meters
-        self.rev_per = self.circum / 299792458  # in seconds
+        self.circum = self.acc.length
```
fernandohds564: I think if the `self.acc` has the attribute length, we don't need this `self.circum` anymore. You can remove it and change the places where it is used.

fernandohds564 (changes requested):

PR #1238, opened 2026-05-08
UE44 IDFF fixes and improvements

ericonr:
We now properly support the multiple tables used by UE44.

### DIFF
--- a/siriuspy/siriuspy/devices/idff.py
+++ b/siriuspy/siriuspy/devices/idff.py
@@ -61,6 +61,15 @@ class _ParamPVs:
                 str_ += lstr if str_ == '' else '\n' + lstr
         return str_
 
+    def _properties_default(self):
+        return tuple(
+            set(
+                value
+                for key, value in _inspect.getmembers(self)
+                if not key.startswith('_') and value is not None
+            )
+        )
+
 
 class IDFFCtrlBase(_Device):
     """ID Feedforward Control Device Base."""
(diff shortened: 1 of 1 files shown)


siriuspy/siriuspy/devices/idff.py
```diff
@@ -449,8 +451,13 @@ def get_ffwd_table_corr_labels(self):
...
         """Return currently active table PV."""
-        # for most hard devices it's just Table-RB
-        return self[self.PARAM_PVS.TABLE_RB]
+        # for hard devices with a single table, it's just that table
+        if not isinstance(self.PARAM_PVS.TABLE_RB, list):
+            return self[self.PARAM_PVS.TABLE_RB]
+
+        # otherwise, we need to choose the table based on the hardware state
+        idx = self[self.PARAM_PVS.TABLEIDX_MON]
+        return self[self.PARAM_PVS.TABLE_RB[idx]]
```
ericonr, line 460: This throws an exception if idx is out of bounds (entirely inconsistent with the device configuration)

Is that ok? It felt like the simplest way of implementing this, instead of checking for inconsistencies here, and propagating that through `get_ffwd_table` below

siriuspy/siriuspy/devices/idff.py
```diff
@@ -61,6 +62,17 @@ def __str__(self):
                 str_ += lstr if str_ == '' else '\n' + lstr
         return str_
 
+    def _properties_default(self):
```
fernandohds564, line 65: I think this should be a @classmethod.

siriuspy/siriuspy/devices/idff.py
fernandohds564: I think it is ok.

siriuspy/siriuspy/devices/idff.py
ericonr: @fernandohds564 , it can't be one (I tested it and it broke the overriding we do).  I've justified it in the commit message:

> Using a @classmethod was considered here, but it doesn't work, since we
override the values of the properties of PARAM_PVS, which is a _ParamPVs
object that belongs to each IDFF class, instead of subclassing
_ParamPVs.

xresende (approved):

siriuspy/siriuspy/devices/idff.py
fernandohds564: ok.

fernandohds564 (approved):

anacso17 (approved):

PR #1032, opened 2023-10-18
Improve ROI methods in DVF device and add more reset parameters

xresende:

PR #1242, opened 2026-05-25
BUG: fix wrong DVF parameters in class constants

xresende:
replacing https://github.com/lnls-sirius/dev-packages/pull/1032

### DIFF
--- a/siriuspy/siriuspy/devices/dvf.py
+++ b/siriuspy/siriuspy/devices/dvf.py
@@ -1,6 +1,7 @@
 """DVF devices."""
 
 import numpy as _np
+
 from mathphys.functions import get_namedtuple as _get_namedtuple
 from mathphys.imgproc import (
     FitGaussianScipy as _FitGaussianScipy,
@@ -25,8 +26,10 @@ class DVF(_Device):
 
     _default_timeout = 10  # [s]
 
+    # should these parameters be moved to csconsts or a DVFSearch?
     _dvfparam_fields = (
-        'MAX_INTENSITY_NR_BITS',
+        'MAX_INTENSITY_NR_BITS',  # [nrbits]
+        'IMAGE_NR_PIXEL_MULTP',  # roi params must be multip. of nr of pixels
         'ACQUISITION_TIME_MIN',  # [s]
         'ACQUISITION_TIME_DEFAULT',  # [s]
         'EXPOSURE_TIME_DEFAULT',  # [s]
@@ -38,19 +41,21 @@ class DVF(_Device):
 
     _dev2params = {
         DEVICES.CAX_DVF1: _get_namedtuple(
+            # Basler acA1300-75gm (version 106755-24)
             'DVFParameters',
             _dvfparam_fields,
-            (16, 0.5, 0.5, 0.100, 2064, 3088, 2.4, 5.0),
+            (16, 16, 0.5, 0.5, 0.100, 2064, 3088, 2.4, 5.0),
         ),
         DEVICES.CAX_DVF2: _get_namedtuple(
             'DVFParameters',
             _dvfparam_fields,
-            (16, 0.5, 0.5, 0.100, 2064, 3088, 2.4, 5.0),
+            (16, 16, 0.5, 0.5, 0.100, 2064, 3088, 2.4, 5.0),
         ),
         DEVICES.BO_DVF: _get_namedtuple(
+            # Basler acA1300-75gm (version 106755-13)
             'DVFParameters',
             _dvfparam_fields,
-            (8, 0.5, 0.5, 0.005, 1024, 1280, 4.8, 5.0),
+            (8, 16, 0.5, 0.5, 0.005, 1024, 1280, 4.8, 5.0),
         ),
     }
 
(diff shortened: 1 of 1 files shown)


siriuspy/siriuspy/devices/dvf.py
henriquesimoes, line 762: I was expecting that the new device specific constant `IMAGE_NR_PIXEL_MULTP` would be used here.

siriuspy/siriuspy/devices/dvf.py
```diff
@@ -38,19 +41,21 @@ class DEVICES:
 
     _dev2params = {
         DEVICES.CAX_DVF1: _get_namedtuple(
+            # Basler acA1300-75gm (version 106755-24)
             'DVFParameters',
             _dvfparam_fields,
-            (16, 0.5, 0.5, 0.100, 2064, 3088, 2.4, 5.0),
+            (16, 16, 0.5, 0.5, 0.100, 2064, 3088, 2.4, 5.0),
         ),
```
henriquesimoes, line 48: As I mentioned [in the deprecated PR](https://github.com/lnls-sirius/dev-packages/pull/1032#discussion_r1400467073), this camera ([acA1300-75gm](https://docs.baslerweb.com/aca1300-75gm)) has a $1024 \times 1280~\text{px}^2$ sensor, not $2064 \times 3088~\text{px}^2$ as defined here.

Can't we always use the existing `cam_max_sizex` and `cam_max_sizey` methods (which read the size of the sensor from the IOC) instead of redefining those constants? As of now, it seems `dvfimgproc/csdev.py` and `dvfimgproc/meas.py` consume these `IMAGE_SIZE_Y` and `IMAGE_SIZE_X` constants defined here.

I just noticed the pixel size is also wrong, because it was replicated from DVF2. It is $4.8\mu m$ for this model. Unfortunately, there is no way of reading this from the camera itself, so my suggestion for the sensor size can't be applied for this.

TL;DR: If we keep redefining the size constants, it should be at least updated to the following:

```suggestion
        DEVICES.CAX_DVF1: _get_namedtuple(
            # Basler acA1300-75gm (version 106755-24)
            'DVFParameters',
            _dvfparam_fields,
            (16, 16, 0.5, 0.5, 0.100, 1024, 1280, 4.8, 5.0),
        ),
```

siriuspy/siriuspy/devices/dvf.py
```diff
@@ -38,19 +41,21 @@ class DEVICES:
...
         DEVICES.CAX_DVF2: _get_namedtuple(
             'DVFParameters',
             _dvfparam_fields,
-            (16, 0.5, 0.5, 0.100, 2064, 3088, 2.4, 5.0),
+            (16, 16, 0.5, 0.5, 0.100, 2064, 3088, 2.4, 5.0),
         ),
```
henriquesimoes, line 53: This camera model is [acA3088-16gm](https://docs.baslerweb.com/aca3088-16gm) (version 107406-12), and its `IMAGE_NR_PIXEL_MULTP` equals to 4, not 16. (Yeah, that's the original one we used for developing the device, and where the constant 4 came from in `cam_roi_calc`).

Thus:

```suggestion
        DEVICES.CAX_DVF2: _get_namedtuple(
            'DVFParameters',
            _dvfparam_fields,
            (16, 4, 0.5, 0.5, 0.100, 2064, 3088, 2.4, 5.0),
        ),
```

siriuspy/siriuspy/devices/dvf.py
```diff
@@ -394,12 +399,20 @@ def cam_frames_failures(self):
 
     def cmd_reset(self, timeout=None):
         """Reset DVF to a standard configuration."""
+        # TODO: is reseting BASLER roi necessary?
```
henriquesimoes, line 402: If you want to ensure it will always be in a valid state, I think so. Otherwise, the client of the `DVF` class will need to know the image might be cut to an arbitrary and possibly irrelevant portion of the sensor.

Why do you think it might not be needed?

siriuspy/siriuspy/devices/dvf.py
```diff
@@ -394,12 +399,20 @@ def cam_frames_failures(self):
...
+        # if not self.cmd_cam_roi_reset():
+        #     return False
         props_values = {
             'cam1:ArrayCallbacks': 1,  # Enable passing array
             'cam1:ImageMode': 2,  # Continuous
             'cam1:PixelFormat': 1,  # Mono12
             # ROI1 takes images from camera driver
             'ROI1:NDArrayPort': self['cam1:PortName_RBV'],
+            # TODO: check if these properties are necessary to be set on reset
+            # 'cam1:DataType': 1,  # UInt16 (maybe unnecessary)
```
henriquesimoes, line 412: Because we set `cam1:PixelFormat` to `Mono12`, the IOC will [automatically update](https://github.com/areaDetector/ADAravis/blob/1455cc46565e2c49ce12d29370ad6253f7e344c0/aravisApp/src/ADAravis.cpp#L801-L813) the `cam1:DataType` to [`UInt16`](https://github.com/areaDetector/ADAravis/blob/1455cc46565e2c49ce12d29370ad6253f7e344c0/aravisApp/src/ADAravis.cpp#L94).

siriuspy/siriuspy/devices/dvf.py
```diff
@@ -394,12 +399,20 @@ def cam_frames_failures(self):
...
+            # 'cam1:NumImages': 1,  # number of sequential acquired images
+            # 'cam1:ExposureMode': 0,  # TIMED
```
henriquesimoes, line 414: It seems our cameras (both models relevant here) don't support any `ExposureMode` other than "timed". So this one in fact isn't currently needed.

siriuspy/siriuspy/devices/dvf.py
```diff
@@ -394,12 +399,20 @@ def cam_frames_failures(self):
...
+        #     return False
         props_values = {
             'cam1:ArrayCallbacks': 1,  # Enable passing array
             'cam1:ImageMode': 2,  # Continuous
             'cam1:PixelFormat': 1,  # Mono12
             # ROI1 takes images from camera driver
             'ROI1:NDArrayPort': self['cam1:PortName_RBV'],
+            # TODO: check if these properties are necessary to be set on reset
+            # 'cam1:DataType': 1,  # UInt16 (maybe unnecessary)
+            # 'cam1:NumImages': 1,  # number of sequential acquired images
```
henriquesimoes, line 413: If we use `ImageMode` as `Continuous`, this PV is irrelevant. Is there any scenario which we use other acquisition mode?

siriuspy/siriuspy/devices/dvf.py
```diff
@@ -394,12 +399,20 @@ def cam_frames_failures(self):
...
+            # 'cam1:ExposureMode': 0,  # TIMED
+            # 'cam1:TriggerMode': 0,  # Off
```
henriquesimoes, line 415: If we ever change the TriggerMode, it seems relevant to ensure this is off. Otherwise, the camera won't acquire because no external trigger will come, right?

I've never configured these cameras to use external trigger, but [it seems that's one of the many parameters required](https://docs.baslerweb.com/triggered-image-acquisition#configuring-triggered-image-acquisition) to use that mode of operation.

henriquesimoes (changes requested):
I noticed this is still a draft, but I'm commenting on everything I noticed nevertheless.

You probably haven't had time to actually implement the use of the new `IMAGE_NR_PIXEL_MULTP` parameter, but I left comment showing where I would expect it to be used.

PR #1248, opened 2026-06-12
Orbit bump device

JeffersonBarrosVieira:
Device created for performing orbit bumps.

The motivation for implementing this device is to avoid repeatedly copying bump implementation logic from scripts. A recurring use case is for XBPM studies.

Must of the logic of orbit bump in the device was taken from the scripts that @Gabrielrezende-asc had developed. Currently, the logic allows performing the bump with SOFB enable. As a future plan, it would be good to generalize te device to be able to perform bumps with FOFB enabled.

On June 8, 2026, @fernandohds564 and I tried to remove some BPMs from the FOFB correction loop. However, when turning on FOFB, FOFB presented errors. So currently the device logic only works with SOFB enable.

Remember that we still need to debug the device.

### DIFF
--- a/siriuspy/siriuspy/devices/__init__.py
+++ b/siriuspy/siriuspy/devices/__init__.py
@@ -54,6 +54,7 @@ from .lienergy import LIEnergy
 from .lillrf import DevLILLRF, LILLRF
 from .machshift import MachShift
 from .modltr import LIModltr
+from .orbit_bump import OrbBump
 from .orbit_interlock import (
     BaseOrbitIntlk,
     BPMOrbitIntlk,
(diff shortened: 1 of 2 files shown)


JeffersonBarrosVieira:
I will close this PR due to the already existing PR in apsuite [add-bumps-to-commisslib](https://github.com/lnls-fac/apsuite/pull/315) which constais classes for performing bumps that @Gabrielrezende-asc and @xresende  were already implementing. I wasn't aware of these ongoing features in apsuite, i apologize fot that. In prniciple, the purpose of performing bumps is the same, however there are already many tools implemented and reviewed in apsuite that would make this redundant here in dev-packages

